PR #46, opened 2023-09-22
#33: improve confirmation modals for vehicle management

radoslavzeman:
- create `useModal` hook for easier work with modals
- cleanup and improve confirmation modals for Add/Delete vehicle actions

Node:
Add vehicle action may not need confirmation modal, but for now let's keep it the same way as Delete action.
Delete modal may be moved to separate screen/component.


app/vehicles/add-vehicle.tsx
```diff
@@ -1,39 +1,53 @@
...
 import { useTranslation } from '@/hooks/useTranslation'
-import { useVehiclesStorage, Vehicle } from '@/hooks/useVehiclesStorage'
+import { useVehicles } from '@/hooks/useVehicles'
+import { Vehicle } from '@/hooks/useVehiclesStorage'
+import { isStandardFormat, sanitizeLicencePlate } from '@/utils/licencePlate'
 
-// TODO make sure that modal is what we want and that it works correctly
 const AddVehicleScreen = () => {
   const t = useTranslation('VehiclesScreen')
+  const tAddModal = useTranslation('VehiclesScreen.addVehicleConfirmModal')
```
mpinter, line 21: not sure if it helps much to structure the translation files and load only parts - in RN, you always have all of them bundled and "downloaded" already anyway. I'd just use translation with 'root namespace and write out the entire path each time the translation is used, making it easy to navigate and change.

Just an opinion, this way is ok as well. Also, if this is done for different reason I'm missing carry on (and let me know).

mpinter (approved):

app/vehicles/add-vehicle.tsx
radoslavzeman: You are right, I wanted to "prepare" it for splitting into separate components, but it's just confusing and hard to read. I'll keep it one level deep in `VehiclesScreen` - that makes sense for me to keep translations grouped "by component" (or related components).

radoslavzeman: I'll use only one `t` function and write longer paths. :)

Issue #57, opened 2023-09-25
SelectRow


radoslavzeman:
![Image](https://github.com/bratislava/paas-mpa/assets/19418224/5fd3abd4-deba-4376-8986-efe51d6ca2d6)

Figma - usage:
- [map](https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-(mobile-app)-%5BWIP%5D?node-id=2677%3A22246&mode=dev)
- [basic single select](https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-(mobile-app)-%5BWIP%5D?node-id=2677%3A22250&mode=dev)
- [multiselect](https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-(mobile-app)-%5BWIP%5D?node-id=3228%3A21265&mode=dev) 

Accessibility issues (from Play Console):
Multiple items have the same description.
This clickable item's speakable text: "" is identical to that of 1 other item(s).

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/f0dc380c-3b21-46bf-8e57-b98d68b82bee)

radoslavzeman:
Component `SelectRow` is created, but it should be revisited and more things should be implemented: 
- checkbox is not scaled properly - this may need custom implementation for checkbox
- whole row should be touchable and should toggle the checkbox
- it should take value from props - this was not tested



![Image](https://github.com/bratislava/paas-mpa/assets/19418224/ba0de3aa-0421-486f-b166-238c1e3279b5)

Issue #56, opened 2023-09-25
ListRow


radoslavzeman:
Figma: https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-(mobile-app)-%5BWIP%5D?node-id=2674%3A21883&mode=dev

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/153e982f-66a3-4576-abc3-4eb8e8e16007)

PR #58, opened 2023-09-25
[#56] [#57] implement ListRow and SelectRow

radoslavzeman:


mpinter (approved):
Yes, would perhaps even consider doing "fake" checkboxes - svg images that look like checked and unchecked checkbox - for this case.

✅

PR #75, opened 2023-09-27
Feature/#22 style nav header

radoslavzeman:
Show `Map` on home screen.
- `Map` is still in `examples` folder and should be moved
- dev menu is available through main menu (or by longPress on main menu button for now - this will be removed)
- map has custom buttons in header area instead of tweaking setting of router header

Header has some basic styles - font and color. Other styles may be added later.

image

#47 
- some missing translations and screens were added into purchase flow

image


mpinter (approved):
needs cleanup of the extra commit of course, but otherwise 🚢

components/controls/vehicles/VehicleRow.tsx
```diff
@@ -1,44 +1,56 @@
...
+  | {
+      onContextMenuPress?: never
+      selected?: boolean
+      showControlChevron?: never
+    }
+  | {
+      onContextMenuPress?: never
+      selected?: never
+      showControlChevron?: boolean
+    }
```
mpinter, line 28: this likely deserves extra comment - for example, I don't know what this achieves - is it that only one of the props can be passed at a time ?

radoslavzeman: Okay, I'll add comment :) And yes, exactly.

PR #76, opened 2023-09-27
Feature/#74 datepicker

radoslavzeman:
Custom DateTime picker implemented.
I took one from "Po meste" project and styled and simplified it for our use-case.
It may need further styling or adjusting, I didn't want to spend much time on it. It seems to work correctly so far.

Note: We may discuss how to pass parameters between screens once again :D

image

### DIFF
--- a/app/purchase/index.tsx
+++ b/app/purchase/index.tsx
@@ -36,8 +36,6 @@ const PurchaseScreen = () => {
   const { getVehicle, defaultVehicle } = useVehicles()
   const { licencePlate, time = '60' } = searchParams
 
-  console.log('searchParams PurchaseScreen:', searchParams)
-
   const chosenVehicle = licencePlate ? getVehicle(licencePlate) : defaultVehicle
 
   // TODO TimeSelector chips sometimes collapses, when not in ScrollView - investigate
--- a/app/purchase/index.tsx
+++ b/app/purchase/index.tsx
@@ -1,10 +1,10 @@
 import BottomSheet, { BottomSheetView } from '@gorhom/bottom-sheet'
-import { Link, router, useLocalSearchParams } from 'expo-router'
+import { Link, router, useGlobalSearchParams } from 'expo-router'
 import React, { useRef } from 'react'
 import { ScrollView } from 'react-native'
 
+import TimeSelector from '@/components/controls/date-time/TimeSelector'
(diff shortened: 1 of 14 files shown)


app/purchase/index.tsx
```diff
@@ -31,8 +32,11 @@ const setTimeValue = (minutes: number) => {
 const PurchaseScreen = () => {
   const t = useTranslation('PurchaseScreen')
   const bottomSheetRef = useRef<BottomSheet>(null)
-  const { licencePlate, time = '60' } = useLocalSearchParams<PurchaseSearchParams>()
+  const searchParams = useGlobalSearchParams<PurchaseSearchParams>()
   const { getVehicle, defaultVehicle } = useVehicles()
+  const { licencePlate, time = '60' } = searchParams
+
+  console.log('searchParams PurchaseScreen:', searchParams)
```
mpinter, line 39: log

mpinter (approved):
Not sure about mixing the local and global search params. Questions:
- choose-vehicle and custom time seem like very similar screensm yet one uses global and one local parameters. Is this a typo ? If it's not , this is quite strange and might be a place for future errors
- I'm not certain about using the global params at all - can we do this without them ?
- it seems we're bringig params to choose-vehicle / custom-time only to then send them back (correct?) - if so, can we do without this ? Somehow merge the params ? Perhaps even - having a local state as the source of truth into which search params are being merged as they are listened for, instead of the params being the source of truth.

---

Skipped over the date-time controls, treating them more or less like copied-over blackboxes.

radoslavzeman:
> Not sure about mixing the local and global search params. Questions:
> 
> * choose-vehicle and custom time seem like very similar screensm yet one uses global and one local parameters. Is this a typo ? If it's not , this is quite strange and might be a place for future errors
> * I'm not certain about using the global params at all - can we do this without them ?
> * it seems we're bringig params to choose-vehicle / custom-time only to then send them back (correct?) - if so, can we do without this ? Somehow merge the params ? Perhaps even - having a local state as the source of truth into which search params are being merged as they are listened for, instead of the params being the source of truth.
> 
> Skipped over the date-time controls, treating them more or less like copied-over blackboxes.

Yeah, I was not sure how to make it work. 
I replaced all Global params by Local params and I'll dig into this state sharing soon.

PR #77, opened 2023-09-27
Feature/#62 determine parking zone

lukybil:
- refactor map code
- show center or map or zoomed out Bratislava
- determine parking zone, the determination is based on the center of the screen and not on coordinates,
  - show price and highlight zone - this was not much additional effort
- added MapPin


modules/map/hooks/useProcessedMapData.ts
```diff
@@ -1,9 +1,9 @@
 import { FeatureCollection } from 'geojson'
 import { useEffect, useState } from 'react'
 
-import { processData } from '@/utils/mapUtils'
+import { useArcgisData } from '@/modules/arcgis/hooks/useArcgisData'
 
-import { useArcgisData } from '../modules/arcgis/hooks/useArcgisData'
+import { processData } from '../utils/processData'
```
radoslavzeman, line 6: please change to absolute path (`@/...`)

modules/map/utils/getIntersectionOfFeatureFromFeatures.ts
```diff
@@ -0,0 +1,32 @@
+import area from '@turf/area'
+import booleanIntersects from '@turf/boolean-intersects'
+import intersect from '@turf/intersect'
+import { Feature, FeatureCollection, Point, Polygon } from 'geojson'
+
+export const getIntersectionOfFeatureFromFeatures = (
```
radoslavzeman: 💔

components/map/Map.tsx
```diff
@@ -15,7 +15,7 @@ import udrStyle from 'utils/layer-styles/visitors2'
 import { useLocation } from '@/modules/map/hooks/useLocation'
 import { useProcessedArcgisData } from '@/modules/map/hooks/useProcessedMapData'
 
-import { MAP_INSETS } from '../constants'
+import { MAP_INSETS } from '../../modules/map/constants'
```
radoslavzeman, line 26: Please use absolute path (`@/...`)

assets/ui-icons/index.ts
```diff
@@ -21,6 +21,7 @@ export { default as BinIcon } from './kos.svg'
 export { default as CrossIcon } from './krizik.svg'
 export { default as SearchIcon } from './lupa.svg'
 export { default as MailIcon } from './mail.svg'
+export { default as MapPin } from './map-pin.svg'
```
radoslavzeman, line 24: `ui-icons` folder is meant just for UI icons (standard 24x24 icons, something like our own version of material icons).
These icons are sed in bratislava design system (bratislava.sk, konto.bratislava.sk...), but finally, we decided to use material icon in Paas MPA, so this folder will probably be removed from repo.

Please create new folder (`map`?) in assets and add pins there (alog with barrel file (index.ts))

components/map/Map.tsx
```diff
@@ -8,14 +8,22 @@ import {
...
   UserTrackingMode,
 } from '@rnmapbox/maps'
-import React, { useRef, useState } from 'react'
+import { MapState, RegionPayload } from '@rnmapbox/maps/lib/typescript/components/MapView'
+import { PermissionStatus } from 'expo-location'
+import { Feature, GeoJsonProperties, Geometry, Point } from 'geojson'
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { View } from 'react-native'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
+import { useDebouncedCallback } from 'use-debounce'
```
radoslavzeman, line 17: Please use `useDebounce` from `usehooks-ts` lib :) 

https://usehooks-ts.com/react-hook/use-debounce

components/map/MapPin.tsx
```diff
@@ -0,0 +1,41 @@
+import { View } from 'react-native'
+
+import MapPinSvg from '@/assets/ui-icons/map-pin.svg'
```
radoslavzeman, line 3: We import svgs as components from barrel files, so they always have same name, and it's easier to work with. This is  just a coding style decision :) 

e.g.
`import { MapPinSvg } from '@/assets/map`

components/map/MapPin.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <View className="absolute items-center" style={position}>
+        {price ? <MapPinSvg {...pinSize} /> : <MapPinNoZoneSvg {...pinSize} />}
+        {price && (
+          <Typography variant="small-bold" className="absolute text-white">
+            {price} €
+          </Typography>
+        )}
+      </View>
+      <View className="absolute" style={{ top: screenCenter.top, left: screenCenter.left }}>
+        <View className="h-[1px] w-[1px] bg-[#f00]" />
```
radoslavzeman, line 36: Just to let you know, that "tailwind values" for 1px size are `-px`, so `h-px w-px` is the syntax that works :)

modules/map/hooks/useScreenCenter.ts
```diff
@@ -0,0 +1,24 @@
+import { useMemo } from 'react'
+import { useWindowDimensions } from 'react-native'
+import { useSafeAreaFrame } from 'react-native-safe-area-context'
+
+interface Props {
```
radoslavzeman, line 5: Another coding style decision is we prefer `types` over `interfaces`. Please change to `type`

modules/map/hooks/useScreenCenter.ts
```diff
@@ -0,0 +1,24 @@
...
+  safeArea?: boolean
+  scale?: boolean
+}
+
+export const useScreenCenter = (props?: Props) => {
```
radoslavzeman, line 10: Name `props` should be used only with components.

Either use two separate arguments  `useScreenCenter(safeArea?: boolean, scale?: boolean)`
or it is also okay to have an "opstions" object as argument, but name it differently.

components/map/MapScreen.tsx
```diff
@@ -33,8 +33,8 @@ const MapScreen = () => {
   return (
     <View className="flex-1 items-stretch">
       <Map onBottomSheetContentChange={handleBottomSheetChange} />
-      {false && (
-        <BottomSheet>
+      {bottomSheetContent && (
+        <BottomSheet snapPoints={['20%', '80%']}>
```
radoslavzeman, line 23: In docs, they use `useMemo` on snapPoints. I don't know yet if it's needed, but just to let you know they use it :D

package.json
```diff
@@ -55,7 +55,8 @@
     "react-native-safe-area-context": "4.6.3",
     "react-native-screens": "~3.22.0",
     "react-native-svg": "~13.9.0",
-    "typescript": "^5.1.3"
+    "typescript": "^5.1.3",
+    "use-debounce": "^9.0.4"
```
radoslavzeman, line 59: Please uninstall if  hook from `usehooks-ts` is used :)

components/map/Map.tsx
```diff
@@ -8,14 +8,15 @@ import {
...
 import { PermissionStatus } from 'expo-location'
-import { Feature, GeoJsonProperties, Geometry, Point } from 'geojson'
+import { Feature, FeatureCollection, GeoJsonProperties, Geometry, Point } from 'geojson'
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { View } from 'react-native'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import { useDebouncedCallback } from 'use-debounce'
-import udrStyle from 'utils/layer-styles/visitors2'
+import udrStyle from 'utils/layer-styles/visitors'
+import udrStyle2 from 'utils/layer-styles/visitors2'
```
radoslavzeman, line 19: Please use absolute paths :)

components/map/Map.tsx
```diff
@@ -115,14 +130,30 @@ const Map = ({ onBottomSheetContentChange }: Props) => {
             animated
           />
         )}
-        {udrData && (
-          <ShapeSource id="udrSource" shape={udrData}>
+        {udrDataByPrice.regular?.features?.length > 0 && (
```
radoslavzeman, line 133: Eslint should complain to this condition, something like "you cannot compare undefined". Isn't it complaining?

radoslavzeman (approved):
Very nice job!

I left several comments, but they are mostly about coding style or thinks that were there also before.

And also, maybe we could move all map related code to `modules` folder? E.g. `utils/layer-styles` :) 

Just one think about eslint and prettier is, that we should have quite good setup now, I spent some time on it again. So no major changes to eslint/prettier setup should be needed. 
Before merging, at least this eslint/prettier setup shoud be reverted. More in slack.

components/map/Map.tsx
radoslavzeman: Or maybe those question marks are not needed?

lukybil: no 😕

lukybil: I reverted the changes to eslint config and now the error is there 👍

.eslintrc.js
```diff
@@ -54,14 +53,8 @@ module.exports = {
...
-    'unicorn/no-array-reduce': 'off',
-    'no-restricted-syntax': 'off',
-    'no-continue': 'off',
+    '@typescript-eslint/no-floating-promises': 'warn',
     'const-case/uppercase': 'off',
-    'prettier/prettier': 'error',
+    'import/extensions': 'off',
   },
-}
+}
```
radoslavzeman, line 60: Turn on this setting in VSCode to keep newlines at the end :) 
```
  "files.insertFinalNewline": true,
```

More in docs: https://bratislava.github.io/onboarding/editor-setup#setup-formatting-and-linting

radoslavzeman (approved):

PR #81, opened 2023-09-28
map icons and bottom sheet

lukybil:
- added map icons as .png images using the `SymbolLayer` which should provide the best performance
  - added simple clustering, styling far from finished
- moved bottom sheet to map, filled with data
  - map pin has yet to be adjusted to be in the middle of the visible map when the bottom sheet is open
- partially fixed types in process data functions
- now centering on the city and not following location if the user is outside the city bounds


assets/map/images/index.ts
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
```
radoslavzeman, line 2: Adding type declaration for png files should help :) (`declarations.d.ts`)
I would try
```
declare module '*.png' {
  import { ImageSourcePropType } from 'react-native'

  const value: ImageSourcePropType;
  export default value;
}
```
https://stackoverflow.com/questions/51100401/typescript-image-import

components/map/Map.tsx
```diff
@@ -9,33 +9,34 @@ import {
...
 
 const Map = ({ onBottomSheetContentChange }: Props) => {
   const camera = useRef<Camera>(null)
   const map = useRef<MapView>(null)
   const [followingUser, setFollowingUser] = useState(true)
-  const { permissionStatus } = useLocationPermission()
+  const [location] = useLocation()
   const insets = useSafeAreaInsets()
-  const screenCenter = useScreenCenter({ scale: Platform.OS === 'android' })
+  const screenCenter = useScreenCenter({ scale: true })
```
radoslavzeman, line 39: I tested it on ios from this branch, it still needs to be `false`

components/map/Map.tsx
```diff
@@ -46,47 +47,48 @@ const Map = ({ onBottomSheetContentChange }: Props) => {
...
-        ...udrData,
-        features: udrData?.features.filter((udr) => udr.properties?.Zakladna_cena === 2),
-      } as FeatureCollection,
-    }),
-    [udrData],
-  )
+  const isWithinCity = useMemo(() => {
+    if (!location) return false
+    const position = [location.coords.longitude, location.coords.latitude]
+    // eslint-disable-next-line sonarjs/prefer-single-boolean-return
```
radoslavzeman, line 81: 👍

modules/map/utils/getMapIcon.ts
```diff
@@ -0,0 +1,32 @@
+import {
+  AssistantIcon,
```
radoslavzeman, line 2: `AssistantImage`?

modules/map/utils/getMapIcon.ts
```diff
@@ -0,0 +1,32 @@
...
+  BranchIcon,
+  GarageIcon,
+  ParkingLotIcon,
+  ParkomatIcon,
+  PartnerIcon,
+  PPlusRIcon,
+} from '@/assets/map'
```
radoslavzeman, line 9: from `@/assests/map/images` ?

lukybil: Oh yes this is an old code, when I was trying to use `.svg`s

components/map/MapScreen.tsx
```diff
@@ -38,13 +39,13 @@ const MapScreen = () => {
...
-            </Field>
-            {bottomSheetContent ? (
-              <>
-                <Panel className="my-2 g-4">
+          <View className="bg-white g-3">
+            <View className="g-2">
+              <Field label={t('MapScreen.BottomSheet.title')}>
+                <TextInput />
+              </Field>
+              {bottomSheetContent && (
```
radoslavzeman: Please prefer ternary operator for non-boolean values such as `bottomSheetContent`. It can lead to UI errors (displaying 0)

e.g. https://dev.to/aishanipach/dont-use-for-conditional-reasoning-3525

radoslavzeman (approved):

components/map/MapScreen.tsx
lukybil: okay will fix in the next PR

PR #82, opened 2023-10-02
Point detail BottomSheet

lukybil:
- added point detail BottomSheet
- points for a discussion:
  - **how to handle point clustering, should all types of points be clustered together?** This is the way it is for the map on the PaaS website, or should the points be clustered separately? This would show the correct icon for clusters, it is the current behavior
  - the data format for the point types is almost completely different for all of them, the variable names are in Slovak (why? 😧) some of them do not have an address, some of them do not have a navigation URL, some do not have a name, **what data do we want to show, since the design handles only one from the 7 cases?**
  - also, the format of the data is very weird to say the least, e.g. `"N/A"` values instead of a `null`, opening hours are not formatted but are a single line of text, etc.
  - **the animation**, right now is done using `LayoutAnimation`, this is the easiest way and looks pretty good, more precise animation requires animating it manually using `Animated` (from `react-native` or `reanimated`)

### DIFF
--- a/components/map/MapMarkers.tsx
+++ b/components/map/MapMarkers.tsx
@@ -11,7 +11,7 @@ import {
   PPLusRImage,
   SellingPointImage,
 } from '@/assets/map/images'
-import { IconsEnum } from '@/modules/map/constants'
+import { MapPointIconEnum } from '@/modules/map/constants'
 import markersStyles from '@/modules/map/utils/layer-styles/markers'
 
 type Props = {
@@ -24,7 +24,7 @@ const MapMarkers = ({ markersData, onPointPress }: Props) => {
     string,
     FeatureCollection<Point, GeoJsonProperties | { icon: string }>,
   ][] = useMemo(() => {
-    return Object.values(IconsEnum).map((icon) => [
+    return Object.values(MapPointIconEnum).map((icon) => [
       icon,
       {
         ...markersData,
@@ -35,8 +35,9 @@ const MapMarkers = ({ markersData, onPointPress }: Props) => {
 
   const handlePointPress = useCallback(
     (event: OnPressEvent) => {
-      if (!event?.features || event.features.length === 0) return
-      onPointPress?.(event.features[0] as Feature<Point, GeoJsonProperties>)
+      if (event?.features && event?.features.length > 0) {
+        onPointPress?.(event.features[0] as Feature<Point, GeoJsonProperties>)
+      }
     },
     [onPointPress],
   )
@@ -45,13 +46,13 @@ const MapMarkers = ({ markersData, onPointPress }: Props) => {
     <>
       <Images
         images={{
-          [IconsEnum.parkomat]: ParkomatImage,
-          [IconsEnum.garage]: GarageImage,
-          [IconsEnum.branch]: SellingPointImage,
-          [IconsEnum.partner]: SellingPointImage,
-          [IconsEnum.pPlusR]: PPLusRImage,
-          [IconsEnum.parkingLot]: ParkingImage,
-          [IconsEnum.assistant]: AssistantImage,
+          [MapPointIconEnum.parkomat]: ParkomatImage,
+          [MapPointIconEnum.garage]: GarageImage,
+          [MapPointIconEnum.branch]: SellingPointImage,
+          [MapPointIconEnum.partner]: SellingPointImage,
+          [MapPointIconEnum.pPlusR]: PPLusRImage,
+          [MapPointIconEnum.parkingLot]: ParkingImage,
+          [MapPointIconEnum.assistant]: AssistantImage,
         }}
       />
       {markersDataByKind?.map(([icon, shape]) => (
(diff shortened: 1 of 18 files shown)


modules/map/types.ts
```diff
@@ -23,6 +25,134 @@ export type SelectedUdrZone = {
...
+  [KindsEnum.partner]: PartnerPoint
+  [KindsEnum.pPlusR]: PPlusRPoinit
+  [KindsEnum.garage]: GaragePoint
+  [KindsEnum.parkingLot]: ParkingLotPoint
+}
 
-  type: 'zone' | 'point'
-} | null
+// Type guard function using conditional types
+export function isPointOfType<T extends KindsEnum>(
```
radoslavzeman, line 153: Type or kind? Pleas unify :)

modules/map/utils/normalizePoint.ts
```diff
@@ -0,0 +1,91 @@
+import i18n from '@/i18n.config'
+
+import { KindsEnum } from '../constants'
```
radoslavzeman, line 3: Please prefer more explicit name - "KindsEnum" - kind of what?
e.g. `MapPointKindEnum` is may be more describing

translations/sk.json
```diff
@@ -83,10 +83,32 @@
...
+      "fields": {
+        "address": "Adresa",
+        "id": "ID",
+        "openingHours": "Otváracie hodiny",
+        "parkingSpotCount": "Počet parkovacích miest",
+        "publicTransportLines": "Linky MHD",
+        "distanceToCenter": "Vzdialenosť do centra",
+        "distanceToPublicTransport": "Vzdialenosť k zastávke MHD"
+      },
+      "kinds": {
```
radoslavzeman, line 103: kinds or types?

components/map/MapMarkers.tsx
```diff
@@ -31,6 +33,14 @@ const MapMarkers = ({ markersData }: Props) => {
     ])
   }, [markersData])
 
+  const handlePointPress = useCallback(
+    (event: OnPressEvent) => {
+      if (!event?.features || event.features.length === 0) return
+      onPointPress?.(event.features[0] as Feature<Point, GeoJsonProperties>)
```
radoslavzeman, line 39: Consistent coding style suggest to write always positive conditions if possible, so:
```
if (event?.features && event?.features.length > 0) {
  onPointPress?.(event.features[0] as Feature<Point, GeoJsonProperties>)
}
```

We also try to limit usage of "if oneliners", because it's easy to miss it while reading the code. It's always better to use braces, even for one-line statement.
```
if (condition) {
   return
}
```

components/map/MapPointBottomSheet.tsx
```diff
@@ -0,0 +1,141 @@
...
+}
+
+const EXCLUDED_ATTRIBUTES = new Set(['address', 'name', 'navigation', 'kind', 'id'])
+
+const MapPointBottomSheet = forwardRef<BottomSheet, Props>(({ point }, ref) => {
+  const t = useTranslation('MapScreen.PointBottomSheet')
+  const [index, setIndex] = useState(-1)
+  const [footerHeight, setFooterHeight] = useState(0)
+
+  const np = useMemo(() => normalizePoint(point), [point])
```
radoslavzeman, line 26: Please don't use abbreviations in variable names, use more descriptive variable name than `np`.

We would usually name it `formattedMapPoint`.

components/map/MapPointBottomSheet.tsx
```diff
@@ -0,0 +1,141 @@
...
+                Object.keys(np)
+                  .filter((k) => !EXCLUDED_ATTRIBUTES.has(k))
+                  .map((k) => (
+                    <View key={k} className="g-4">
+                      <Divider />
+                      <Field label={t(`fields.${k}`)}>
+                        <Typography>{np[k as keyof typeof np]}</Typography>
+                      </Field>
+                    </View>
+                  ))}
```
radoslavzeman, line 132: This could be written by ternary operator.

components/map/MapPointBottomSheet.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  const handleClose = useCallback(() => {
+    localRef.current?.close()
+  }, [])
+
+  useEffect(() => {
+    localRef.current?.collapse()
+  }, [point])
+
+  const expanded = index === 1
```
radoslavzeman, line 73: We try to prefix boolean variables with `is` - `isExpanded`.

radoslavzeman (commented):
~~When I click on Parkomat icon, a get~~
Sometimes the click on map point show this error:
```
 ERROR  Error: Text strings must be rendered within a <Text> component.
```

components/shared/Field.tsx
```diff
@@ -1,21 +1,22 @@
...
   label: string
   labelInsertArea?: ReactNode
   children: ReactNode
   errorMessage?: string
+  variant?: TypographyProps['variant']
 }
 // TODO associate control with label
 
-const Field = ({ label, children, labelInsertArea, errorMessage }: Props) => {
+const Field = ({ label, children, labelInsertArea, errorMessage, variant }: Props) => {
```
radoslavzeman, line 15: Please set default variant here, not when used (line 19)

lukybil: oh yes a good catch

radoslavzeman (approved):

PR #84, opened 2023-10-02
init openapi generated axiosClient

radoslavzeman:
Initiate generated axios client.

Update aplify libs.

TODO next: Use token in interceptor.


app/examples/modal.tsx
```diff
@@ -30,6 +30,7 @@ const ModalScreen = () => {
           title="Confirm modal"
           text="Optional text"
           primaryActionLabel="Some action"
+          // eslint-disable-next-line no-console
```
lukybil, line 33: we are using the console logs in so many places and do not disable the warning

lukybil (approved):

app/examples/modal.tsx
radoslavzeman: Yes, we should disable the warning on on those places where console log is intentional and permanent.
When console log is just temporary, we should keep the warning on, for future reference.

app/examples/healthcheck.tsx
```diff
@@ -0,0 +1,33 @@
...
+import Typography from '@/components/shared/Typography'
+import { clientApi } from '@/modules/backend/client-api'
+
+const HealthcheckScreen = () => {
+  const [healthcheck, setHealthcheck] = useState<boolean | null>(null)
+
+  clientApi
+    .appControllerHealth()
+    .then((res) => {
+      if (res.status) setHealthcheck(true)
```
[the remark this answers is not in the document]
radoslavzeman: Exactly :)

Issue #65, opened 2023-09-25
Map - Zone detail screen


radoslavzeman:
Figma: https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-(mobile-app)-%5BWIP%5D?node-id=1828%3A15506&mode=dev

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/dd0f1eb6-dda8-4c36-946a-97df8488ec62)

PR #85, opened 2023-10-04
#65: Zone detail screen

lukybil:
- Zone detail screen
- New way of passing zone and point data between screens using global store
- type names cleanup


app/zone-details/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+const ZoneDetailsScreen = () => {
+  // reason for not passing the whole object: https://reactnavigation.org/docs/params/#what-should-be-in-params
+  const zoneDetailsParams = useLocalSearchParams<ZoneDetailsParamas>()
+  const t = useTranslation()
+
+  const id = zoneDetailsParams.id ? Number.parseInt(zoneDetailsParams.id, 10) : -1
+
+  const zone = useMapZone(id, true)
+
+  if (!zone) {
```
lukybil, line 26: I know that it is a negative condition, but it saves indentation and makes it clearer what happens, for return value of Elements, I would do it this way

app/zone-details/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+const ZoneDetailsScreen = () => {
+  const zoneDetailsParams = useLocalSearchParams<ZoneDetailsParamas>()
+  const globalStore = useContext(GlobalStoreContext)
+
+  const id = Number.parseInt(zoneDetailsParams.id ?? '0', 10)
+
+  const zone = useMemo(
+    () => globalStore.mapFeatures?.get(id) as Feature<Polygon, SelectedUdrZone>,
```
radoslavzeman: it would be safer without casting, if possible

state/GlobalStoreProvider.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+type ContextProps = {
+  mapFeatures: MapFeatureHashMap | null
+  setMapFeatures: Dispatch<SetStateAction<MapFeatureHashMap | null>>
+}
+
+export const GlobalStoreContext = createContext({} as ContextProps)
+
+type Props = {
+  children: React.ReactNode
```
radoslavzeman, line 18: We write just `ReactNode` and import it.

state/GlobalStoreProvider.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const GlobalStoreProvider = ({ children }: Props) => {
```
radoslavzeman, line 21: But you can also use just `PropsWithChildren` from react.

app/zone-details/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { SelectedUdrZone } from '@/modules/map/types'
+import { GlobalStoreContext } from '@/state/GlobalStoreProvider'
+
+type ZoneDetailsParamas = {
+  id: string
+}
+
+const ZoneDetailsScreen = () => {
+  const zoneDetailsParams = useLocalSearchParams<ZoneDetailsParamas>()
+  const globalStore = useContext(GlobalStoreContext)
```
radoslavzeman: Please create `useGeneralStoreContext` for this :)

radoslavzeman (approved):
Nice :)

modules/map/hooks/useNormalizedPoint.ts
```diff
@@ -0,0 +1,19 @@
...
+import { NormalizedPoint, SelectedPoint } from '@/modules/map/types'
+import { normalizePoint } from '@/modules/map/utils/normalizePoint'
+
+export function useNormalizedPoint<Point extends SelectedPoint | undefined | null>(
+  point: Point,
+): Point extends undefined | null ? NormalizedPoint | null : NormalizedPoint
+export function useNormalizedPoint(
+  point: SelectedPoint | undefined | null,
+): NormalizedPoint | null {
+  const [, i18n] = useTranslation()
```
radoslavzeman, line 13: You can also use `useLocale()` custom hook we have: `const locale = useLocale()`. This is also how other intl libraries work. 
Or if you need i18n instance, we usually do `const { i18n } = useTranslation()`

state/hooks/useMapZone.ts
```diff
@@ -1,29 +1,36 @@
...
+  | (NormalizePropertiesOnly extends true ? NormalizedUdrZone : Feature<Polygon, SelectedUdrZone>)
+  | null
 export function useMapZone(
   id: number,
-  propertiesOnly: boolean = false,
-): SelectedUdrZone | Feature<Polygon, SelectedUdrZone> | null {
+  normalizedPropertiesOnly: boolean = false,
+): NormalizedUdrZone | Feature<Polygon, SelectedUdrZone> | null {
   const globalStore = useContext(GlobalStoreContext)
+  const [, i18n] = useTranslation()
```
radoslavzeman, line 18: useLocale

app/zone-details/index.tsx
```diff
@@ -19,7 +19,7 @@ const ZoneDetailsScreen = () => {
   const zoneDetailsParams = useLocalSearchParams<ZoneDetailsParamas>()
   const t = useTranslation()
 
-  const id = Number.parseInt(zoneDetailsParams.id ?? '0', 10)
+  const id = zoneDetailsParams.id ? Number.parseInt(zoneDetailsParams.id, 10) : -1
```
radoslavzeman, line 22: null instead of -1?

radoslavzeman (approved):

PR #90, opened 2023-10-05
feature/more components

radoslavzeman:
see commits


app/settings/language.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+// TODO
+const Page = () => {
+  const t = useTranslationLocal('Settings')
+  const { i18n } = useTranslation()
+  const [mmkvLocale] = useMMKVString(STORAGE_LANGUAGE_KEY)
+
+  const setLanguage = async (language: 'sk' | 'en') => {
+    await i18n.changeLanguage(language)
+
```
lukybil, line 26: Doesn't the `mmkvLocale` have to be updated? The language changes, but the UI shows always only "System"

lukybil (approved):
Tested on Android

app/settings/language.tsx
radoslavzeman: Yes, thanks!

Issue #98, opened 2023-10-11
Switch component


radoslavzeman:
![Image](https://github.com/bratislava/paas-mpa/assets/19418224/e6b20d10-0457-41c2-abbf-6633e4c2fca9)

Used in Settings: https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-mpa?node-id=486%3A2961&mode=dev

Original figma from DS ESBS: https://www.figma.com/file/17wbd0MDQcMW9NbXl6UPs8/DS-ESBS%2BBK%3A-Component-library?node-id=12%3A439&mode=dev

PR #100, opened 2023-10-12
Feature/#99-notification-settings

radoslavzeman:
- setup react query lib
- implement simple Switch component - potentially will be replaced by styled variant #98 
- implement requests to get settings and change settings
- show notification settings on Settings page

TODO : #101 

Figma: https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-mpa?node-id=1321%3A10432&mode=dev

### DIFF
--- a/components/controls/notifications/NotificationSettings.tsx
+++ b/components/controls/notifications/NotificationSettings.tsx
@@ -1,0 +1,92 @@
+import { useQuery } from '@tanstack/react-query'
+
+import NotificationControl from '@/components/controls/notifications/NotificationControl'
+import Field from '@/components/shared/Field'
+import { useTranslation } from '@/hooks/useTranslation'
+import { clientApi } from '@/modules/backend/client-api'
+import { SaveUserSettingsDto } from '@/modules/backend/openapi-generated'
+
+const NotificationSettings = () => {
+  const t = useTranslation('Settings')
+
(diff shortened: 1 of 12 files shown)


components/controls/notifications/NotificationSettings.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  // TODO catch error
+  const saveNotifications = async (value: SaveUserSettingsDto) => {
+    try {
+      await clientApi.usersControllerSaveUserSettings(value)
+      await refetch()
+    } catch (error) {
+      console.log(error)
+    }
+  }
```
lukybil, line 29: When I switch the switch e.g. from off to on, it firstly turns to on for a brief moment then off again and then finally on, this is probably caused by the fact that the switches are directly bound to the server data
I would use the "Optimistic updates" pattern here and add an intermediate front end state layer, the switch would directly switch to the desired value, and we would only handle an erroneous response (a mismatch between the backend and frontend data)

lukybil (approved):
The comment can be solved in a future PR

components/controls/notifications/NotificationSettings.tsx
radoslavzeman: Definitely, creating new issue for this.

PR #110, opened 2023-10-13
create ticketPriceRequest state in global store

radoslavzeman:


components/tickets/PurchaseBottomSheet.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+type Props = object
+
+const PurchaseBottomSheet = forwardRef<BottomSheet, Props>((props, ref) => {
+  const t = useTranslation('PurchaseScreen')
+
+  const { ticketPriceRequest } = useGlobalStoreContext()
+
+  const insets = useSafeAreaInsets()
+  const purchaseButtonContainerHeight = 24 + 12 + 48 + insets.bottom
```
lukybil, line 22: I'm not quite sure what exactly those numbers are, please maybe add a comment or extract it to some constant

lukybil (approved):
🙌

components/tickets/PurchaseBottomSheet.tsx
radoslavzeman: Oh yes, sorry, just fixed "pixels" from Figma, but I think we want to handle this some better way...

PR #111, opened 2023-10-14
Full height search

lukybil:
- search now extends to a full screen `BottomSheet`
- all is kept in a single `BottomSheet`, it provides a better user experience but was difficult to get right and it still has not been yet tested on  iOS
- basic ideas about the functionality:
  - the full height can only be accessed after pressing the `TextInput`, it cannot be manually extended
  - after it is extended it can be closed either by swiping, pressing the cancel button or choosing an option
  - after selecting an option, the input field displays the option data
  - when the input field is pressed again, it reverts to the last user input, this is there because the full address is too long, when the user would want to delete it or edit their last search
- `LayoutAnimation` is used for the very nice transitions between states
- the animations can sometimes be a little weird, but this can be fixed
- the largest problem is that the Map reacts to the Keyboard and resizes, changing the selected zone, this is for now resolved by ignoring the changes when the bottom sheet is fully extended
- I would like to collect feedback on the functionality


components/shared/Field.tsx
```diff
@@ -20,7 +21,7 @@ const Field = ({
   variant = 'default-bold',
 }: Props) => {
   return (
-    <View className="g-1">
+    <View className={clsx('g-1', children === null && 'pb-1')}>
```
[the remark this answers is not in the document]
radoslavzeman: Do you need to use `Field` for that case?

radoslavzeman (commented):
I was not able to open the fullscreen search. A simple press on search field does nothing.

components/inputs/TextInput.tsx
```diff
@@ -1,28 +1,43 @@
...
+  ({ hasError, isDisabled, leftIcon, ...rest }, ref) => {
+    const localRef = useRef<TextInputNative>(null)
+    const refSetter = useMultipleRefsSetter(localRef, ref)
+
+    const handlePress = useCallback(() => {
+      localRef.current?.focus()
+    }, [])
+
+    return (
+      <TouchableWithoutFeedback
```
radoslavzeman, line 28: I don't like that some basic component should depend on "a random" library that has nothing to do with text inputs or styling.

components/shared/Field.tsx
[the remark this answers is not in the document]
lukybil: I want to use the style of the `Field` but want to keep the element outside

components/inputs/TextInput.tsx
```diff
@@ -1,28 +1,48 @@
...
+    return (
+      <TouchableWithoutFeedback onPress={handlePress}>
+        <View
+          className={clsx('flex-row rounded border bg-white px-4 py-3 g-3', {
+            'border-divider focus:border-dark': !isDisabled && !hasError,
+            'border-negative': hasError && !isDisabled,
+            'border-divider bg-[#D6D6D6]': isDisabled,
+          })}
+        >
+          <View className="h-6 w-6">{leftIcon ?? null}</View>
```
radoslavzeman, line 40: Using ternary operator will prevent to render empty View:
```
{leftIcon ? <View className="h-6 w-6">{leftIcon}</View> : null}
```

That's how we usually handle conditional rendering.

radoslavzeman: Please also provide `aria-hidden` prop to icon view, so it's hidden for screen readers. (at least that's what would I do in web)

components/map/MapAutocomplete.tsx
```diff
@@ -1,29 +1,60 @@
...
-    <Autocomplete
-      getOptions={forwardGeocode}
-      getOptionLabel={(option) => option.place_name || option.text}
-      onValueChange={handleValueChange}
-    />
-  )
-}
+    const renderItem: ListRenderItem<GeocodingFeature> = useCallback(
+      ({ item }) => (
+        <View className="border-b-[2px] border-divider py-4">
```
radoslavzeman, line 27: 2px is default :) 
image

components/inputs/TextInput.tsx
radoslavzeman: The first comment will probably solve also this new empty space on all text inputs:
image

components/inputs/Autocomplete.tsx
```diff
@@ -32,13 +52,23 @@ const AutocompleteInner = <O,>(
...
+    resultsHeader,
+    optionsPortalName,
+    ListComponent = FlatList,
+    listProps = {},
+    ...restProps
+  }: AutocompleteProps<O>,
+  ref: React.ForwardedRef<RNTextInput>,
 ) => {
   const [input, setInput] = useState(defaultValue)
   const [value, setValue] = useState<O | null>(null)
```
radoslavzeman, line 68: This `value` seem not to be used anywhere.

components/shared/Field.tsx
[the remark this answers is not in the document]
radoslavzeman: I would rather manually set the same style on "Kde parkujete" label, than make this component ugly for one usecase :D

radoslavzeman (approved):

Issue #102, opened 2023-10-13
Use native datetime picker


radoslavzeman:
Use native datetime picker instead of custom one.

It shouldn't be possible to choose datetime in past. The way how to ensure this is up to developer.

https://github.com/henninghall/react-native-date-picker

PR #112, opened 2023-10-15
#102 Native datetime picker

lukybil:
- replaced the custom datetime picker with a native one from `react-native-date-picker`
- date picker in modal mode


hooks/useTranslation.ts
```diff
@@ -14,5 +14,8 @@ export const useTranslation = (keyPrefix?: string) => {
 export const useLocale = () => {
   const { i18n } = useTranslationI18n()
 
-  return i18n.language
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [locale, region] = i18n.language.split('-')
+
+  return locale as 'sk' | 'en'
```
radoslavzeman, line 20: I would always return what `i18n.language` returns.
If there is "long version" like `sk-SK` or `en-UK`, it should be investigated where this happens and it should be converted there.
For datetime, we should manually chose UK, US, or IE version of dates. Preferably UK, but for some reason we have IE date formats on our web projects.

I'm open to discussion :)

Issue #106, opened 2023-10-13
Request: Parking cards (verify email)


radoslavzeman:
Figma: https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-mpa?node-id=4232%3A5935&mode=dev

Issue #103, opened 2023-10-13
Optimistic update for Notification settings


radoslavzeman:
(On android) When I switch the Switch e.g. from off to on, it firstly turns to on for a brief moment then off again and then finally on, this is probably caused by the fact that the switches are directly bound to the server data.

Use the "Optimistic updates" pattern here and add an intermediate front end state layer, the Switch would directly switch to the desired value, and we would only handle an erroneous response (a mismatch between the backend and frontend data)

Try: https://tanstack.com/query/latest/docs/react/guides/optimistic-updates

radoslavzeman:
I tried to implement optimistic update following react query docs, but it still seems to not work on android.
https://github.com/bratislava/paas-mpa/pull/122

PR #122, opened 2023-10-19
feat: #106 request parking cards

radoslavzeman:
- upgrade React Query to new v5.0.0
- fix `Button`'s disabled state on loading
- add some translations
- make some eslint rules only warn - these may potentially be disabled completely, because they are triggered by other libraries such as react query, which we do not control 
- sent request with new email to verify (token should be returned for now) - TODO mailgun does not work now
- implement optimistic update for Notification settings  - cloces #103 

TODO
- error handling - show snackbar?
- typing for optimistic update
- etc..

### DIFF
--- a/components/controls/notifications/NotificationSettings.tsx
+++ b/components/controls/notifications/NotificationSettings.tsx
@@ -1,4 +1,4 @@
-import { useQuery } from '@tanstack/react-query'
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 
 import NotificationControl from '@/components/controls/notifications/NotificationControl'
 import Field from '@/components/shared/Field'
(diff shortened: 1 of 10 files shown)


components/controls/notifications/NotificationSettings.tsx
```diff
@@ -60,31 +119,26 @@ const NotificationSettings = () => {
...
-          <NotificationControl
-            key={setting.name}
-            notificationName={setting.name}
-            value={setting.value}
-            onValueChange={() =>
-              saveNotifications({
-                ...notificationSettings,
-                [setting.name]: !setting.value,
-              })
-            }
```
lukybil, line 84: The switches sadly have the same behavior as before, switching for a brief moment, then switching back, before finally switching to the desired position, I think this is a problem with the Switch itself, that it expects to be always switched after a press. The solution would be to use a state for the switches that updates instantly, because the mutation seems to be a little bit too slow, or change it for a custom Switch, which would not be difficult

lukybil (approved):
Very cool! I love the mutations and looking at the code it should work perfectly but it looks like that there is a problem with toggable controls 🙃

PR #123, opened 2023-10-20
Map functionalities and active tickets

lukybil:
- added `BottomSheetTopAttachment` component for attaching nodes above the bottom sheet, implemented it for the zone bottom sheet - active tickets, gps go to location button
- improved Map:
  - removed scale
  - added compass, can be changed for a custom image
  - now blocking zone updates when Keyboard is present (because the Map weirdly shrinks), this solves other problems as well, like bugs with the zone bottom sheet
- refactored large `Map` component


components/map/Map.tsx
```diff
@@ -77,29 +76,9 @@ const Map = forwardRef(
     const { markersData, zonesData, udrData, odpData } = useFilteredMapData(processedData, filters)
 
     useEffect(() => {
-      onZoneChange?.((selectedPolygon?.properties as MapUdrZone) ?? null)
+      onZoneChange?.(selectedPolygon ? selectedPolygon?.properties : null)
```
radoslavzeman, line 79: This does not work?
```js
onZoneChange?.(selectedPolygon?.properties ?? null)
```

components/map/MapZoneBottomSheetAttachment.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <BottomSheetTopAttachment {...restProps}>
+      <FlexRow cn="flex-1 p-[10px] pt-0 items-end">
+        <View className="rounded-full bg-white shadow ">
+          <Link asChild href="/tickets">
+            <PressableStyled>
+              <FlexRow cn="g-2 p-2 pr-3 items-center">
+                <View className="h-8 w-8 items-center justify-center rounded-full bg-light">
+                  <Icon size={20} name="local-parking" />
+                </View>
+                <Typography variant="default-bold" className="py-1">
```
radoslavzeman, line 43: There is no `py-1` in figma

components/map/MapZoneBottomSheetAttachment.tsx
```diff
@@ -0,0 +1,64 @@
...
+      if (!location) {
+        location = await Location.getCurrentPositionAsync()
+      }
+      setFlyToCenter?.([location.coords.longitude, location.coords.latitude])
+    }
+  }, [setFlyToCenter, permissionStatus])
+
+  return (
+    <BottomSheetTopAttachment {...restProps}>
+      <FlexRow cn="flex-1 p-[10px] pt-0 items-end">
```
radoslavzeman, line 35: `p-[10px] pt-0` --> `px-2.5 pb-2.5`

I hadn some issues with arbitrary values that they didn't scale to `dp` units. I think it's safer to use `-2.5` that we have defined in tailwind config.

radoslavzeman (approved):

components/map/Map.tsx
```diff
@@ -58,11 +58,10 @@ const Map = forwardRef(
...
     const screenCenter = useScreenCenter({ scale: Platform.OS === 'android' })
-    const [selectedPolygon, setSelectedPolygon] = useState<Feature<
-      Geometry,
-      GeoJsonProperties
-    > | null>(null)
-    const [selectedPoint, setMapInterestPoint] = useState<Feature<Point, GeoJsonProperties> | null>(
+    const [selectedPolygon, setSelectedPolygon] = useState<Feature<Polygon, MapUdrZone> | null>(
+      null,
+    )
+    const [selectedPoint, setMapInterestPoint] = useState<Feature<Point, MapInterestPoint> | null>(
```
radoslavzeman, line 64: `selectedPoint` is not used, also some constants are left on top. Consider removing the disabled rule `@typescript-eslint/no-unused-vars` and enable it again :)

components/map/Map.tsx
lukybil: it was causing problems with the type assertion but now its removed so I will change it back to `??`

components/map/MapZoneBottomSheetAttachment.tsx
lukybil: It looked like that `p-[10px]` was different than `p-2.5` when I used it but this was probably caused by the hot reload and Map being present

PR #125, opened 2023-10-22
Get and display active tickets

lukybil:
- tickets page display active tickets
- no active tickets screen
- terminate ticket modal
- using test data


modules/backend/hooks/useTickets.ts
```diff
@@ -0,0 +1,61 @@
...
+      createdAt: '2022-01-03T10:00:00Z',
+    },
+  ],
+}
+
+export const useTickets = (active: boolean) => {
+  const options = useMemo(
+    () =>
+      queryOptions({
+        queryKey: [`tickets-${active ? 'active' : 'past'}`],
```
radoslavzeman, line 50: It's completely okay to pass "active/past" as second item into the array.

assets/avatars/sad-smiley.svg
radoslavzeman, line 1: We realised, that it's the best to use the same name of the file as in Figma.

components/map/MapZoneBottomSheetAttachment.tsx
```diff
@@ -12,13 +12,16 @@ import Icon from '@/components/shared/Icon'
...
+import { useTranslation } from '@/hooks/useTranslation'
+import { useTickets } from '@/modules/backend/hooks/useTickets'
 import { useLocationPermission } from '@/modules/map/hooks/useLocationPermission'
 
 type Props = Omit<BottomSheetTopAttachmentProps, 'children'> & {
   setFlyToCenter?: MapRef['setFlyToCenter']
 }
 
 const MapZoneBottomSheetAttachment = ({ setFlyToCenter, ...restProps }: Props) => {
+  const t = useTranslation('MapScreen.ZoneBottomSheet.TopAttachment')
```
radoslavzeman, line 24: Here the same, ZoneBottomSheet doesn't have to be nested in MapScreen.

state/hooks/useMapZone.ts
```diff
@@ -5,32 +5,31 @@ import { useLocale } from '@/hooks/useTranslation'
...
 export function useMapZone(
-  id: number | null,
+  udrId: number | string | null,
   normalizedPropertiesOnly: boolean = false,
 ): NormalizedUdrZone | Feature<Polygon, MapUdrZone> | null {
   const globalStore = useGlobalStoreContext()
   const locale = useLocale()
 
   return useMemo(() => {
+    const id = typeof udrId === 'string' ? Number.parseInt(udrId, 10) : udrId
```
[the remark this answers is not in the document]
lukybil: It has been then found that udr id is a number

radoslavzeman (approved):

components/map/MapZoneBottomSheetAttachment.tsx
```diff
@@ -30,23 +33,30 @@ const MapZoneBottomSheetAttachment = ({ setFlyToCenter, ...restProps }: Props) =
...
-                  <Icon size={20} name="local-parking" />
-                </View>
-                <Typography variant="default-bold" className="leading-6">
-                  Tickets (2)
-                </Typography>
-              </FlexRow>
-            </PressableStyled>
-          </Link>
-        </View>
+        {activeTicketsCount > 0 ? (
```
radoslavzeman, line 42: Our practice is usually to check it like this. You have to use ternary in this case always - `undefined`, `null` or `0` evaluates to false, that is what we want usually:
```
tickets?.length ? <Component/> : null
```

And you don't have to create unnecessary variable :)

components/map/MapZoneBottomSheetAttachment.tsx
```diff
@@ -30,23 +33,30 @@ const MapZoneBottomSheetAttachment = ({ setFlyToCenter, ...restProps }: Props) =
...
+                  </View>
+                  <Typography variant="default-bold" className="leading-6">
+                    {t('tickets', { count: activeTicketsCount })}
+                  </Typography>
+                </FlexRow>
+              </PressableStyled>
+            </Link>
+          </View>
+        ) : (
+          <View />
```
radoslavzeman, line 58: Is this empty View what we want?

PR #146, opened 2023-10-31
feat/#144: implement verification result screens

radoslavzeman:
- implement `/parking-cards/verification/verification-result` page, that handle redirects from BE to show email verification status
- refetch verified emails when added or deleted
- reorganize verification screens
- add some TextInput props to improve UX
- add empty state screens
  - EmptyStateScreen still need some updates (it's not always whole screen, reflect screen title and content title...)

TODO
- deduplicate mutation code for sending verification email


lukybil (approved):
On Android I get redirerected to browser from the link in the email. But that is problably not a part of this PR, so looks good 👍

PR #150, opened 2023-11-03
#107: Announcements

lukybil:
- added Announcements screen
- for now with mocked data, ready for backend data
- TODO: How do we translate the announcements? Send a param with locale to the backend?


corrections.d.ts
radoslavzeman, line 1: This file may need to be added to `tsconfig.ts`.
Please add a comment with link to docs or an issue, why this is needed.

Consider more specific filename, such as `markdown.d.ts`.

components/info/SeveritySquare.tsx
```diff
@@ -0,0 +1,41 @@
+import { clsx } from 'clsx'
+import { View } from 'react-native'
+
+import Icon, { IconName } from '@/components/shared/Icon'
+
+type Props = {
+  variant: 'info' | 'warning' | 'danger'
```
radoslavzeman, line 7: Is there a reason to use different variant names than defined in swagger?

radoslavzeman (approved):

PR #151, opened 2023-11-03
fix registration process to updated setup in Cognito

radoslavzeman:
Changes in Cognito:
Registration (= sign up) does not send verification sms.
User is signed up and confirmed (confirmation is set in custom `Pre sign-up Lambda trigger`).
Phone number is not verified automatically, but this is okay, because phone is automatically verified during login.

- after successful sign up, we trigger login and redirect to confirm-login screen.
- move `confirmSignIn` to hook
- remove unused empty interceptors
- show snackbar for sign up/ sign in errors
- do not prefetch when not signed in

TODO
- handle errors properly
- document the whole sign up/sign in flow


hooks/useSignInOrSignUp.ts
```diff
@@ -1,55 +1,83 @@
...
+    } catch (error) {
+      // TODO
+      // [NotAuthorizedException: Invalid session for the user, session is expired.]
+      // [CodeMismatchException: Invalid code or auth state for the user.]
+      if (isErrorWithCode(error) && error.code === 'CodeMismatchException') {
+        console.log('Code mismatch', error)
+        snackbar.show(`Code mismatch: ${error.message}`, { variant: 'warning' })
+      } else if (isError(error)) {
+        console.log('Confirm Error', error)
+        snackbar.show(`Confirm Error: ${error.message}`, { variant: 'danger' })
```
lukybil, line 71: Translations 😆

lukybil (approved):
Tested on Android and it works very well ✅

PR #152, opened 2023-11-06
Feat/#16 code input

radoslavzeman:
- implement `CodeInput` based on `react-native-confirmation-code-field`
- use `CodeInput` on `confirm-signin` screen
- regenerate api and update FE accordingly
- use `CodeInput` in email verification process

Note
- It would be probably better to implement inner state by some `useControlledState` hook, instead of forcing us to pass `value` and `setValue` props, but we can implement it if it causes some issues.
- `CodeInput` needs to be tested properly, mainly on smaller screen.
- Whole email verification flow will be updated when figma is ready for verification with code instead of link
  - We still display verification code also on FE, for easier development

image

### DIFF
--- a/app/confirm-signin.tsx
+++ b/app/confirm-signin.tsx
@@ -1,6 +1,6 @@
 import { useState } from 'react'
 
-import TextInput from '@/components/inputs/TextInput'
+import CodeInput from '@/components/inputs/CodeInput'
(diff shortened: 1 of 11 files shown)


lukybil (commented):
Tested on Android, very nicely done!

PR #164, opened 2023-11-08
Fix: quickfix EmptyStateScreen + minor updates

radoslavzeman:
- conditionally render `<Stack.Screen>` only if `title` is provided - this allows us to use `EmptyStateScreen` inside `ScreenView` without overriding title
- few minor changes in `EmptyStateScreen`
- add some translations for Tickets screen
- update api


lukybil (approved):
✅

PR #168, opened 2023-11-10
Feat/#167 new payment options

radoslavzeman:
This PR implements new payment options - Apple Pay and Google Pay

- update Avatar components
- use more suitable names for components
  - "PaymentOptions" for payment-card/apple-pay/google-pay
  - "PaymentMethods" for PaymentOptions + visitor cards
- implement default payments option with mmkv storage - TODO this may need some improvements
- implement screen for setting default payment option (before it was the same as choosing payment method)


lukybil (approved):
Looking good 👍

PR #185, opened 2023-11-20
Improvement project structure and contexts

PavelFilo:
### **File structure changes:**
To have protected routes we need to have a global _layout file and one inside (app) folder to route back to the sign screen. This approach is based on expo-router documentation.

### **Context changes:** 
1. changes to have optimized context providers with less rerenders and some unnecessary useEffect removals.
2. Made PurchaseStoreContext available to map page so when I want to change place the filled fields won't change.

### **Layout fixes:**
 Purchase routes stack was separated so expo-router did not give button to go back to the map from the purchase index route.


hooks/useVehicles.ts
```diff
@@ -21,7 +21,7 @@ export const useVehicles = () => {
     }
   }
 
-  const defaultVehicle = vehicles.length > 0 ? vehicles[0] : null
+  const defaultVehicle = vehicles?.[0] || null
```
radoslavzeman, line 24: Use `??` to prevent `undefined`. Or if it's okay to have `undefined` as vehicle, `??/|| null` does not make sense at all.

radoslavzeman (commented):
Nice job!

I would be useful to add a link to documentation that you mentioned in PR decsription. Somewhere in ts docs into main layout for example.

radoslavzeman (approved):

hooks/useVehicles.ts
PavelFilo: undefined is prevented as you can see in the screenshot with `||` syntax as well
Screenshot 2023-11-23 at 14 01 09

radoslavzeman: yes, sorry, you are correct

radoslavzeman (approved):

Issue #177, opened 2023-11-13
Empty vehicles state


radoslavzeman:
Show proper Empty state instead of temporary one.

Current temporary solution:

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/fbc611b1-6be8-4948-94a0-d0a39b16db93)

Issue #156, opened 2023-11-07
BottomSheet - add shadow


radoslavzeman:
https://gorhom.github.io/react-native-bottom-sheet/adding-shadow

lukybil:
I would use the `react-native-shadow` library that we already include, because the Android's elevation shadow does not really provide a top shadow

Issue #129, opened 2023-10-25
Loading states


radoslavzeman:

lukybil:
We could use skeletons in some places

radoslavzeman:
As we use prefetch almost everywhere, so I would implement just simple spinner.

PR #199, opened 2023-11-26
Sprint 11 smaller tasks

lukybil:
- #177 
- #156 added shadows to BottomSheets that do not have a backdrop
- #129 added loading spinner to components that do not have skeletons, added loading behavior to the `MapZoneBottomSheet`
- changed the behavior of the `MapZoneBottomSheet` to extend on pan up
- please test the loading spinner on iOS

### DIFF
--- a/components/map/MapZoneBottomSheet.tsx
+++ b/components/map/MapZoneBottomSheet.tsx
@@ -70,18 +70,16 @@ const MapZoneBottomSheet = forwardRef<BottomSheet, Props>((props, ref) => {
     if (isZoneSelected) {
       newSnapPoints.push(SNAP_POINTS.zone + bottom)
     }
-    if (isFullHeightEnabled) {
-      newSnapPoints.push(SNAP_POINTS.searchExpanded)
-    }
+    newSnapPoints.push(SNAP_POINTS.searchExpanded)
 
     return newSnapPoints
-  }, [isZoneSelected, isFullHeightEnabled, bottom])
+  }, [isZoneSelected, bottom])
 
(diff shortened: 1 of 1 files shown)


app/(app)/vehicles/index.tsx
```diff
@@ -61,19 +61,23 @@ const VehiclesScreen = () => {
     closeModal()
   }
 
+  if (vehicles.length === 0) {
+    return (
+      <ScreenView title={t('title')} backgroundVariant="dots">
+        <NoVehicles />
+      </ScreenView>
```
radoslavzeman, line 68: Meh, I don't like repeating the title like this, but I know we don't have better approach now. :D

components/screen-layout/LoadingScreen.tsx
```diff
@@ -0,0 +1,31 @@
...
+import ScreenView from '@/components/screen-layout/ScreenView'
+
+type LoadingScreenProps = {
+  asScreenView?: boolean
+} & ViewProps
+
+const LoadingScreen = ({ asScreenView, className, ...rest }: LoadingScreenProps) => {
+  if (asScreenView) {
+    return (
+      <ScreenView
```
radoslavzeman, line 13: Maybe `ScreenViewCentered` is what you want?

radoslavzeman (approved):

radoslavzeman:
I have some kind of "border" on top of BottomSheet. Is this what we want?
image

radoslavzeman:
Tested on ios, loading spinner works well.

PR #201, opened 2023-11-27
Feature error messages

PavelFilo:
- Error handling for the price calculation in the bottom sheet of the purchase screen.
- Pricing bottom sheet height improvements to be able to change the height based on its content.
- improvements to not show all the errors in snackbar.
- removal of eslint rule `exhaustive-deps` because not all dependencies should be in dependencies everytime!


components/info/ErrorMessage.tsx
```diff
@@ -0,0 +1,10 @@
+import clsx from 'clsx'
+
+import Panel from '@/components/shared/Panel'
+import Typography from '@/components/shared/Typography'
+
+export const ErrorMessage = ({ message, className }: { className?: string; message: string }) => (
```
radoslavzeman, line 6: Please follow consistent code style (https://bratislava.github.io/frontend/components):
- default export instead of named export
- props type as separate type (usually called `ErrorMessagePanelProps`, or just `Props`, if not exported)
- more explicit component name, such as `ErrorMessagePanel`

components/tickets/PurchaseBottomSheet.tsx
```diff
@@ -153,15 +166,24 @@ const PurchaseBottomSheet = forwardRef<BottomSheet, Props>(
...
+            setPurchaseButtonContainerHeight(event.nativeEvent.layout.height)
+          }}
+        >
           {/* Toggling visibility instead hiding by "display: none" prevents layout shifts */}
-          <FlexRow className={priceData ? 'visible' : 'invisible'}>
+
+          <FlexRow className={priceData ? 'visible' : 'hidden'}>
             <Typography variant="default-bold">{t('summary')}</Typography>
-            {priceData ? (
+            {priceData && (
```
radoslavzeman, line 180: https://dev.to/maafaishal/avoid-operator-for-conditional-rendering-in-react-2de

modules/backend/hooks/useAxiosResponseInterceptors.ts
```diff
@@ -8,55 +8,53 @@ import { axiosInstance } from '@/modules/backend/axios-instance'
...
-      }
 
-      snackbarMessage ??= t('generic')
+          snackbarMessage ??= t('axiosGeneric')
+          break
+      }
+    } else snackbarMessage ??= t('generic')
 
-      snackbar.show(snackbarMessage, { variant: 'danger' })
+    if (snackbarMessage) snackbar.show(snackbarMessage, { variant: 'danger' })
```
radoslavzeman, line 47: Please use code-block braces after if statement. We also want to force it be eslint soon.

.eslintrc.js
```diff
@@ -68,5 +68,7 @@ module.exports = {
     'sonarjs/cognitive-complexity': 'warn',
     '@typescript-eslint/no-misused-promises': 'warn',
     '@typescript-eslint/ban-ts-comment': 'warn',
+    // many times you dont want to add everything to deps to avoid unnecessary triggers of hooks. for example useEffect.
+    'react-hooks/exhaustive-deps': 'off',
```
radoslavzeman, line 72: We definitely want this rule enabled. If needed, you can always disabled in locally.

app/(app)/(purchase-and-payment)/purchase/index.tsx
```diff
@@ -28,12 +27,19 @@ export type PurchaseSearchParams = {
   udrId?: string
 }
 
+interface IPriceError {
```
radoslavzeman, line 30: We prefer types to interfaces. Use interface only if type cannot be used.
https://bratislava.github.io/frontend/components

modules/backend/hooks/useAxiosResponseInterceptors.ts
```diff
@@ -8,55 +8,53 @@ import { axiosInstance } from '@/modules/backend/axios-instance'
...
+  const errorInterceptor = (error: unknown) => {
+    let snackbarMessage = null
+    if (isAxiosError(error)) {
+      const { status, data } = error.response ?? {}
+      const { errorName, message }: { errorName?: string; message?: string } = data
+
+      // eslint-disable-next-line sonarjs/no-small-switch
+      switch (status) {
+        case 422:
+          break
```
radoslavzeman, line 29: Please add docs in comment, why we skip this types of error (where they should be handled etc...)

modules/backend/hooks/useAxiosResponseInterceptors.ts
```diff
@@ -8,55 +8,53 @@ import { axiosInstance } from '@/modules/backend/axios-instance'
 
 // https://dev.to/arianhamdi/react-hooks-in-axios-interceptors-3e1h
 
+const successInterceptor = (response: AxiosResponse) => {
+  return response
+}
```
radoslavzeman, line 13: I would keep both interceptors together inside `useAxiosResponseInterceptors` function and disable the eslint rule that forces us to put it outside locally.

app/(app)/(purchase-and-payment)/purchase/index.tsx
radoslavzeman: Is PurchaseScreen the right place to store this type/interface?

radoslavzeman (commented):
For the bottom sheet size, there is `enableDynamicSizing` prop, but we found out today, that there is some bug with it. I would not introduce custom approach for sizing if it is not really needed. If it is, please add a comment to code what problem it is solving.
If it's solving some open issue, it should also be marked as TODO.

This is the issue we found today https://github.com/gorhom/react-native-bottom-sheet/issues/1573
We will address it very soon.

PavelFilo:
> For the bottom sheet size, there is `enableDynamicSizing` prop, but we found out today, that there is some bug with it. I would not introduce custom approach for sizing if it is not really needed. If it is, please add a comment to code what problem it is solving.

I've added `TODO:` into code for this, because the bottom sheet height needs to be looked at with its drawing functionality

radoslavzeman (approved):

Issue #218, opened 2023-11-30
Display "zoom in" info in MapZoneBottom


radoslavzeman:
Display "zoom in" info in MapZoneBottom while MapPin is not displayed.

https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-mpa?type=design&node-id=2493-18888&mode=design&t=8MRK0B3ptc3YJ6oh-0

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/0dbc31fe-d2ba-4bde-b75d-bf66fa2db937)

PR #224, opened 2023-11-30
#223: Map zone bottom sheet cleanup

lukybil:
- `MapZoneBottomSheet` cleanup, fixed some bugs #223 
- added zoom in text when map pin not shown #218 
- changed to static arcgis data source


components/map/MapZoneBottomSheet.tsx
```diff
@@ -112,139 +80,60 @@ const MapZoneBottomSheet = forwardRef<BottomSheet, Props>((props, ref) => {
...
+                {...{
+                  ref: inputRef,
+                  handleInputBlur,
+                  isFullHeight,
+                  setIsFullHeight,
+                  selectedZone,
+                  setFlyToCenter,
+                  bottomSheetRef: localRef.current,
+                }}
+              />
```
radoslavzeman, line 133: I would prefer standard props passing, or potentially, create separate const to be spread here (meybe without refs?).

radoslavzeman (approved):
TBH, it's still hard to understand what's going on in here, even tho I know how it should work.
Ideally, the code should be easily understandable for any developer, even for juniors. But maybe this is just complicated use-case that will always take time to read through.

But is's definitely nice step further :)

Screenshot from ios - zoomed out bottom sheet is quite low on screen:
image

PR #228, opened 2023-12-04
Feature vehicles api

### LINKED ISSUE #226: Rewrite Vehicles to use API endpoints instead
Currently, vehicles are stored in app storage and API endpoints are not used at all. Reason: we implemented vehicles as very first feature, when API was not ready yet.

We want to use API endpoints instead, so users will have their saved vehicles also when changing phone device.

If API does not suite our use-cases, please consult with @pkulcsarnr, it can be changed according to needs.

Image

PavelFilo:
closes #226 

- vehicles queries and mutations,
- vehicles store instead of only hook,
- loading when adding vehicle
- purchase context licencePlate string change for vehicle object
- naming in vehicle workflow


state/VehiclesStoreProvider/VehiclesStoreProvider.tsx
radoslavzeman, line 1: Please use `queryOptions` and then `someOptions.().queryKey` instead of duplicating the key everywhere. Put the the query next to others in `queryOptions.ts` file.

radoslavzeman (changes requested):
It's definitely a good step forward.

I think that there are cases that should be handled on BE, for example BE should provide us with the list of vehicle where there is always one and only one default vehicle, or empty list. Please consult with @pkulcsarnr about possible changes in BE.

I was also not able to add vehicle without a name. See video attached: I also had an issue with first vehicle, that it didn't appear in the list on first render, but I was not able to capture it on video.

https://github.com/bratislava/paas-mpa/assets/19418224/c84d53bc-5db3-43f7-a548-fabd37cc4524

app/(app)/(purchase-and-payment)/purchase/index.tsx
```diff
@@ -57,17 +56,17 @@ const PurchaseScreen = () => {
     ticket: {
       udr: String(udr?.udrId) ?? '',
       udrUuid: udr?.udrUuid ?? '',
-      ecv: licencePlate || '',
+      ecv: licencePlate ?? '',
```
radoslavzeman, line 59: `?? ''` is not needed when you added it above

modules/backend/constants/queryOptions.ts
```diff
@@ -178,5 +178,4 @@ export const vehiclesOptions = () =>
   queryOptions({
     queryKey: ['Vehicles'],
     queryFn: () => clientApi.vehiclesControllerVehiclesGetMany(),
-    select: (res) => res.data,
   })
```
radoslavzeman, line 181: I would maybe add a comment why you don't use `select: (res) => res.data`, otherwise, I think people will try to implement it :D

radoslavzeman (approved):
Nice!

Please add a link to docs for the cache update as we discussed.

app/(app)/vehicles/index.tsx
```diff
@@ -41,8 +46,8 @@ const VehiclesScreen = () => {
     [],
   )
 
-  const handleContextMenuPress = (licencePlate: string) => {
-    setActiveVehicle(licencePlate)
+  const handleContextMenuPress = (id: number) => {
```
radoslavzeman, line 49: Maybe `id` --> `vehicleId` to make it more specific?

PR #248, opened 2023-12-05
displaying confirm-sign-in errors

PavelFilo:


components/inputs/CodeInput.tsx
```diff
@@ -40,44 +41,50 @@ type CodeInputProps = Omit<
...
+                'border-negative': !!error,
+              },
+            )}
+          >
+            {symbol || (isFocused ? <Cursor /> : null)}
+          </Typography>
+        )}
+      />
+
+      {error ? <Typography className="text-negative">{error}</Typography> : null}
```
radoslavzeman, line 86: If `react-native-confirmation-code-field` does not have an error props, than okay.
But we should keep in mind also accessibility.

app/(auth)/confirm-sign-in.tsx
```diff
@@ -35,17 +47,53 @@ const Page = () => {
     }
   }
 
+  const handleInputFocus = () => {
+    if (errorCode) {
+      setErrorCode('')
+    }
+  }
+
+  const handleConfirmSignIn = async () => {
```
radoslavzeman, line 56: Trigger this only if code length is valid, because it's triggered on blur.

radoslavzeman (approved):

radoslavzeman (approved):

components/inputs/CodeInput.tsx
PavelFilo: There is no single line about errors in that lib... I even tried to make it myself, but as I read in react-native docs there is no such thing as `aria-invalid` or any similar accessibility field...

PR #262, opened 2023-12-07
Feature delete account

PavelFilo:


app/(app)/settings/index.tsx
```diff
@@ -32,6 +35,16 @@ const SettingsPage = () => {
     [],
   )
 
+  const { mutate, isPending } = useMutation({
+    mutationFn: () => clientApi.usersControllerDeleteUser(),
+    onSuccess: async (response) => {
+      if (response.data) {
+        const responsee = await Auth.deleteUser()
+        if (responsee === 'SUCCESS') await signOut()
```
radoslavzeman, line 43: Please use braces

app/(app)/settings/index.tsx
```diff
@@ -32,6 +35,16 @@ const SettingsPage = () => {
     [],
   )
 
+  const { mutate, isPending } = useMutation({
```
radoslavzeman, line 38: I would follow the style as on other mutations, to use whole mutation object instead of destructuring. What do you think?

app/(app)/settings/index.tsx
```diff
@@ -92,10 +101,11 @@ const SettingsPage = () => {
       <Modal visible={isModalVisible} onRequestClose={toggleModal}>
         <ModalContentWithActions
           variant="error"
+          isLoading={isPending}
           title={t('deleteAccountConfirmModal.title')}
           text={t('deleteAccountConfirmModal.message')}
           primaryActionLabel={t('deleteAccountConfirmModal.actionConfirm')}
-          primaryActionOnPress={handleConfirmDelete}
+          primaryActionOnPress={mutate}
```
radoslavzeman, line 108: If used like this, I would like to see some more specific name than `mutate`. e.g. `mutateSignOut`?

app/(app)/settings/index.tsx
```diff
@@ -32,6 +35,16 @@ const SettingsPage = () => {
     [],
   )
 
+  const { mutate, isPending } = useMutation({
+    mutationFn: () => clientApi.usersControllerDeleteUser(),
+    onSuccess: async (response) => {
+      if (response.data) {
+        const responsee = await Auth.deleteUser()
```
radoslavzeman, line 42: `responsee` - double "e" ?

There is CodeSpell Checker plugin to VS Code :)  https://bratislava.github.io/onboarding/editor-setup#extensions

app/(app)/settings/index.tsx
radoslavzeman: As I understand, we still have to implement deleting on Cognito? Please add a TODO comment.

radoslavzeman (approved):

app/(app)/settings/index.tsx
PavelFilo: that was just for testing purposes and I didn't remove it (response was already defined two lines above 😄 ). But thanks for suggestion :)

app/(app)/settings/index.tsx
radoslavzeman: Ah, yes, I just didn't see it :D

Issue #241, opened 2023-12-05
MainMenu as screen with transparent background


radoslavzeman:

radoslavzeman:
Implemented as standard screen with animation from left, for simplicity.

Issue #231, opened 2023-12-04
MapZoneBottomSheet text ovelap


radoslavzeman:
![Image](https://github.com/bratislava/paas-mpa/assets/19418224/c718c668-3aea-4153-9a31-d648b20c4d3f)

Issue #220, opened 2023-11-30
Newlines in markdown doesn't work


radoslavzeman:
![Image](https://github.com/bratislava/paas-mpa/assets/19418224/a756bd9e-9ea8-4ee6-b190-053253e66186)

PR #298, opened 2023-12-17
Main menu screen & bug fixes

lukybil:
- resolved comments from the previous PR (I did not change the SearchContext because I did not want to end up with a function `setSetFlyToCenter`, I can update it if necessary)
- #220 newlines in markdown fix
- #272 navigate to current location button: If permission is denied, the button is hidden, if location service is off, after pressing the button, location should be requested to be turned on
- #241 main menu as screen
- #231 text overlap fix

Points for a discussion:
- Should the main menu have a fixed distance?
- What should be the size of the spacer on the left? Right now the menu has a min width and is set to flex grow
- Should the spacer be a backdrop? Should it have a lets say `rgba(0,0,0,0.5)` background?


PavelFilo (approved):

components/shared/Markdown.tsx
```diff
@@ -41,7 +61,11 @@ type Props = {
 const Markdown = ({ children, fontSize = 'default', textCenter }: Props) => {
   const rules = useMemo(() => getRules(fontSize, textCenter), [fontSize, textCenter])
 
-  return <MarkdownNative rules={rules}>{children}</MarkdownNative>
+  return (
+    <MarkdownNative rules={rules} markdownit={markdownInstance} debugPrintTree>
```
radoslavzeman, line 65: Do we want `debugPrintTree` for prod?

radoslavzeman (approved):
Please double check all `nearBy`/`NearBy` occurrences and use `b` instead of `B` :D

radoslavzeman:
Questions can be addressed in Figma and solved as separate PR.

radoslavzeman:
Safe view should be used in multiple places, it's a good point, but I would keep it as separate issue.

Issue #295, opened 2023-12-13
MapPin wrongly indicates zone on ios


radoslavzeman:
It happens on ios, and it's probably caused by querying rect instead of point. (?)

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/530db24f-5d8d-45ed-952a-acca5a1dee0c)

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/16f9905d-26f5-47ea-8049-75c2a0b96e5f)

Issue #293, opened 2023-12-13
Address search result redundant text


radoslavzeman:
Address points in "Address" search results show redundant text, see screenshot:

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/b804fa16-e512-4f61-a520-46c77a92826d)

PR #322, opened 2023-12-21
Sprint 15 map bug fixes

lukybil:
- #295 reduce rectangle size, introduce screen scale, must be tested on iOS
- #283 disable 3D map view
- #259 change `MapPointBottomSheet` title to the kind of the point
- #260 `MapPointBottomSheet` change the shown attributes
- #293 remove redundant address text


components/map/MapPointBottomSheet.tsx
```diff
@@ -71,8 +92,11 @@ const MapPointBottomSheet = forwardRef<BottomSheet, Props>(({ point }, ref) => {
     localRef.current?.collapse()
   }, [point])
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
PavelFilo, line 95: why? :) just remove it if not used

PavelFilo (approved):

PR #336, opened 2023-12-27
Improvement ticket shortening

### LINKED ISSUE #119: Request: Terminate ticket
Terminating is possible with `/shorten/{id}`

- The minimum price for a ticket is 5 cents so it will cancel just to that point, no further in case of immediate cancellation after purchase.
- ticket can be terminated only before `parkingEnd` time, otherwise throws an error.
- prolonged tickets can be canceled just to the last prolongation time start, not to the initial ticket time.
- all the possible errors should be displayed just as toasts in snackbar for now (Zuzka made a note about error displaying while shortening).

PavelFilo:
closes #119


app/(app)/tickets/shorten/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { ticketId } = useLocalSearchParams<{ ticketId: string }>()
+
+  if (!ticketId) {
+    router.replace('/tickets')
+  }
+
+  const shortenTicketMutation = useMutation({
+    mutationFn: (id: number) => clientApi.ticketsControllerShortenTicket(id),
+    onSuccess: () => {
+      queryClient.removeQueries({ queryKey: ['Tickets'] })
```
lukybil, line 26: The `queryKey` should be taken from the `activeTicketsOptions`, so that we do not lose track of where the keys are used.

app/(app)/tickets/shorten/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+              <View className="grow">
+                <Markdown>{t('infoText')}</Markdown>
+
+                <View className="mt-5 g-1">
+                  <Typography variant="default-bold" className="mb-1">
+                    {t('mainConditions')}
+                  </Typography>
+
+                  {(t('conditions', { returnObjects: true }) as unknown as string[]).map((item) => (
+                    <View className="max-w-full flex-row g-2">
```
lukybil, line 59: a `key` prop is missing

lukybil (approved):
👍

app/(app)/tickets/shorten/index.tsx
PavelFilo: This is different, because I want to remove all the queries that start with `"Tickets"` key, not just one query

PavelFilo:
Please test all the headers in the app if they have the correct title after these changes (I found one that hasn't, fixed it, and pushed it to this PR)

Issue #337, opened 2023-12-27
After adding an email, navigate to emails list


lukybil:
Right now the user is navigated to add parking cards screen with prefilled email, instead of getting to the emails list

Issue #313, opened 2023-12-21
Header alignment on android


radoslavzeman:
Our custom header back button is misaligned on android. It works nicely on ios.

Please center the title horizontally (from left to center) - this is default on ios but not on android.
Try to center the back button and title also vertically.

On Menu screen, please remove X button.

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/e4ab2b71-0be8-4f69-8bbf-ab040bc08563)

Issue #188, opened 2023-11-24
Search - version 1.0.2.


Zanarova:
Testfligt build 1.

Steps to reproduce: 
- write adress or code
- search results are shown
- scroll down
- headlines (parkovacie zóny, adresy) are not fixed, they override result while scrolling

Expected result: 
- fix  headlines (parkovacie zóny, adresy)

![Image](https://github.com/bratislava/paas-mpa/assets/108303461/941f92eb-580f-4599-add0-dd22970fce0d)

Zanarova:
![Image](https://github.com/bratislava/paas-mpa/assets/108303461/941f92eb-580f-4599-add0-dd22970fce0d)

PR #339, opened 2023-12-28
Sprint 16 bug fixes

lukybil:
- #188 disable sticky headers
- #271 render user location after other shapes
- #324 delete assistants from the whole app
- #313 The positioning in the middle was not easy because of the left arrow, the title was not centered, this could be solved with an absolutely posisitoned container, but this solution works without absolutely positoned elements
- #337 changed one `router.push` to `router.replace`


app/(app)/parking-cards/verification/index.tsx
```diff
@@ -31,7 +31,7 @@ const Page = () => {
     onSuccess: (res) => {
       const tmpVerificationToken = res.data[0].token
 
-      router.push({
+      router.replace({
```
PavelFilo, line 34: cannot be done like this, because if user wants to go back from verification-sent to email it wont be possible, just leave it with `push` and I will adjust new part of code I added to verification-result page :)

components/screen-layout/StackScreenWithHeader.tsx
```diff
@@ -1,27 +1,83 @@
...
+              />
+            ) : null}
+            <View
+              className="min-w-0 flex-1 flex-row items-center justify-center"
+              style={{
+                marginLeft: headerSideWidth - headerLeftWidth,
+                marginRight: headerSideWidth - headerRightWidth,
+              }}
+            >
+              <Typography variant="h1" className="bottom-[3px] max-w-full text-ellipsis">
```
PavelFilo, line 50: font size of header title is too big, compare it with design

PavelFilo (changes requested):
Make sure the header is done according to the design :)

PavelFilo:
Please test all the headers in the app if they have the correct title after these changes (I found one that hasn't, fixed it, and pushed it to this PR)

app/(app)/parking-cards/verification/index.tsx
PavelFilo: done here: https://github.com/bratislava/paas-mpa/pull/340

lukybil:
> Please test all the headers in the app if they have the correct title after these changes (I found one that hasn't, fixed it, and pushed it to this PR)

What do you mean by this please? I changed back the title that you changed. It should be passed to the `ScreenView` to render the custom title.

app/(app)/(purchase-and-payment)/purchase/choose-vehicle.tsx
```diff
@@ -53,10 +53,9 @@ const ChooseVehicleScreen = () => {
   }
 
   return (
-    <ScreenView>
+    <ScreenView title={t('title')}>
```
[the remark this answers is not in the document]
lukybil: Of course I have tested it. Aren't there also two `Stack.Screen` for all those screens where we set other attributes like animation or whether it's a modal in layout routes?

PR #340, opened 2023-12-28
navigate back to correct route from validation result page

### LINKED ISSUE #103: Optimistic update for Notification settings

### LINKED ISSUE #337: After adding an email, navigate to emails list

PavelFilo:
- fixed going back to emails page from validation success page (closes #337)
- optimistic updates on notificationSettings (closes #103)

### DIFF
--- a/app/(app)/parking-cards/verification/verification-result.tsx
+++ b/app/(app)/parking-cards/verification/verification-result.tsx
@@ -52,12 +52,19 @@ const VerificationResultPage = () => {
     if (status === 'link-expired') return
 
     const state = navigation.getState()
-    const index = state.routes.findIndex((route) => (route.name as string).includes('purchase'))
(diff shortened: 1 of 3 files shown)


lukybil (approved):
Could not test it because of the block by park dots but the code looks correct 👍

PR #354, opened 2024-01-10
#321: Gis data aliases

lukybil:
- change gis data to our internal format in `processData`
- use only our internal format in the rest of the application
  - removed the initial elegant `Proxy` solution, because the translated objects should be reactive to the language of the app
- add weekends and holidays price, and labels to zone detail
- added support for aliased gis data, for now supporting both aliased and not-aliased data formats


modules/map/utils/processData.ts
```diff
@@ -117,92 +124,109 @@ export const processData = ({
...
+          .filter((f) => f.properties?.web === 'ano')
           .map((feature) => {
             GLOBAL_ID++
             const type =
-              feature.properties?.Typ_en == 'P+R'
-                ? 'p-plus-r'
-                : feature.properties?.Typ_en == 'garage'
-                ? 'garage'
-                : 'parking-lot'
+              (feature.properties as Arcgis.ParkingPoint)?.Typ_en == 'P+R' ||
```
PavelFilo, line 200: same here, why only 2? how can there be different type?

modules/map/utils/translateMapObject.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function translateMapObject<
+  P extends MapPoint | MapUdrZone,
+  K extends keyof P,
+  T extends WithTranslationProperties<P, K>,
+>(obj: T, locale: ApplicationLocale): P {
+  const translatedObject: P = {} as P
+  Object.entries(obj).forEach(([key, value]) => {
+    translatedObject[key as keyof P] =
+      typeof value === 'object' && value !== null
```
PavelFilo, line 29: i think that `typeof value === 'object' && value` does the same thing

PavelFilo (approved):

PR #362, opened 2024-01-12
#345: Weekend ticket prices

lukybil:
- display weekend and holidays price

### DIFF
--- a/app/(app)/zone-details.tsx
+++ b/app/(app)/zone-details.tsx
@@ -31,8 +31,7 @@ const ZoneDetailsScreen = () => {
 
   const price = getPriceFromZone(zone)
 
-  // eslint-disable-next-line eqeqeq
-  const isAllWeekPrice = zone.weekendsAndHolidaysPrice == undefined
+  const isAllWeekPrice = zone.weekendsAndHolidaysPrice == null
 
   return (
     <ScreenView title={t('title')}>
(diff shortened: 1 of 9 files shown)


app/(app)/zone-details.tsx
```diff
@@ -28,13 +29,18 @@ const ZoneDetailsScreen = () => {
     return null
   }
 
+  const price = getPriceFromZone(zone)
+
+  // eslint-disable-next-line eqeqeq
+  const isAllWeekPrice = zone.weekendsAndHolidaysPrice == undefined
```
PavelFilo, line 35: why only `==`?

utils/getPriceFromZone.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const holidays = new Holidays('SK')
+
+export const getPriceFromZone = (
+  zone: MapUdrZone | MapUdrZoneWithTranslationProps,
+  date?: Date,
+) => {
+  const localDate = date ?? new Date()
+  // eslint-disable-next-line eqeqeq
+  if (zone.weekendsAndHolidaysPrice == undefined) {
```
PavelFilo, line 13: same here, why?

app/(app)/zone-details.tsx
```diff
@@ -57,11 +63,20 @@ const ZoneDetailsScreen = () => {
           </FlexRow>
         ) : null}
         {/* eslint-disable-next-line eqeqeq */}
+        {isAllWeekPrice || zone.price == undefined ? null : (
```
PavelFilo, line 66: also here... if you want to check if its null or undefined create a function or add there both undefined and null with ===

PavelFilo (approved):

app/(app)/zone-details.tsx
lukybil: changed to `== null` since this is also allowed by eslint

radoslavzeman: Last time we discussed this in the office, we agreed that `== null` is hardly readable a prone to bugs, so it's better to use `isDefined` function, that we have in almost all projects.
This is just a note for future :)

PR #389, opened 2024-01-22
debouncing duration

### LINKED ISSUE #357: Debounce for price calculation

PavelFilo:
- added debounce for the duration in price calculation (both prolongation and first purchase) (closes #357)
the first purchase has to have a separate function to have the freshest parkingStart when clicking the pay button
prolongation doesn't have this usecase because using parkingStart from an old ticket and not the current time


radoslavzeman (approved):
LGTM.

PR #392, opened 2024-01-22
Fix/ios gpay

radoslavzeman:
- hide GPay on ios
- fix translations on email verification screen
- hide NPK original balance temporarily


PavelFilo (approved):
nice.

PR #408, opened 2024-01-29
add `ticketId` to redirect on android

radoslavzeman:
This is just quickfix. I think we should treat the `ticketId` somehow more reliable, not just search params where it can get lost/forgotten.


PavelFilo (approved):
nice

Issue #366, opened 2024-01-15
Define accessibility tasks


radoslavzeman:
Using Google Play console.

PR #411, opened 2024-02-02
Accessibility changes based on iOS VoiceOver testing

PavelFilo:
changes found in voiceover testing like: 
- DismissKeyboard component accessibility turned to false so it wont show in reader
- old Keyboard.dismiss removal or swap with DismissKeyboard
- AccessibilityField correcting label 
- translation for a place where no Field component is used.
linked issue - #366

### DIFF
--- a/components/map/MapZoneBottomSheet.tsx
+++ b/components/map/MapZoneBottomSheet.tsx
@@ -64,6 +64,7 @@ const MapZoneBottomSheet = forwardRef<BottomSheet, Props>((props, ref) => {
                   <TextInput
                     pointerEvents="none"
                     ref={inputRef}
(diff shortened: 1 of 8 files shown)


components/map/MapZoneBottomSheet.tsx
```diff
@@ -64,6 +64,7 @@ const MapZoneBottomSheet = forwardRef<BottomSheet, Props>((props, ref) => {
                   <TextInput
                     pointerEvents="none"
                     ref={inputRef}
+                    accessibilityLabel={t('searchAccessibilityInput')}
```
radoslavzeman, line 67: I guess it would be better to use `AccessibilityField`.

radoslavzeman (approved):

PR #427, opened 2024-02-11
onboarding skip button move to top

PavelFilo:


app/onboarding.tsx
```diff
@@ -124,7 +124,15 @@ const OnboardingScreen = () => {
             accessibilityLabel={t('goBack')}
             name="arrow-back"
           />
-        ) : null}
+        ) : (
+          <View />
```
radoslavzeman, line 128: ?

radoslavzeman (approved):
Are we sure about empty View?

app/onboarding.tsx
PavelFilo: 100% :D there are 2 ways to do it, one is class and the second is empty view :) this was a bit easier in my opinion

PR #455, opened 2024-03-11
HERE WE GO - expo updates

### LINKED ISSUE #300: Use `eas update`

PavelFilo:
Long awaited feature is finally here! EXPO-UPDATES

- nativewind upgrade to v4
- fixes of structural changes in nativewind version like class ordering or deprecated classes like `divider`
- added tailwind-merge library and clsx custom function to use by default
- bottom sheet dynamic sizing fix instead of fixed min-height px
- text input focus as a state because new nativewind doesn't apply focus on input parent
- added important to tailwind classes because of icons that had stronger color classes (text-[color] did nothing without important keyword)
- map performance IMPORTANT fix!! the rotation coused query refetch with each degree of rotation... caused immediate `too many requests` error by API and app was unusable,
- closes #300 

needs testing of whole app both on iOS and Android


components/tickets/PurchaseBottomSheet.tsx
```diff
@@ -39,9 +39,13 @@ const PurchaseBottomSheet = forwardRef<BottomSheet, Props>(
...
       >
-        <BottomSheetContent cn="g-3" hideSpacer>
+        {/**
+         * Better approach for zero height bottom sheet: https://github.com/gorhom/react-native-bottom-sheet/issues/1573
+         * fixes dynamic height of bottom sheet
+         */}
+        <BottomSheetScrollView scrollEnabled={false}>
           {priceData ? (
-            <>
+            <View className="gap-3 px-5 py-3">
```
radoslavzeman, line 48: `gap-3` --> `g-3`?

utils/clsx.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const twMerge = extendTailwindMerge({
+  extend: {
+    classGroups: {
+      'font-size': ['text-h1', 'text-h2', 'text-h3', 'text-base', 'text-sm'],
+    },
+  },
+})
+
+export const clsx = (...args: cn.ClassValue[]) => twMerge(cn(...args))
```
radoslavzeman, line 12: I would go with name `cn` instead of `clsx`, similar as `shadcn` does it.

radoslavzeman (approved):

PR #466, opened 2024-04-09
feat: editing name of vehicle

### LINKED ISSUE #462: Edit vehicle

PavelFilo:
- functionality to edit created vehicle name (closes #462)


app/(app)/vehicles/edit-vehicle.tsx
```diff
@@ -0,0 +1,86 @@
...
+  if (!vehicleId) return null
+
+  const handleSaveVehicle = async () => {
+    if (!vehicle) return
+
+    await editVehicle({
+      id: +vehicleId,
+      licencePlate: vehicle.vehiclePlateNumber,
+      vehicleName,
+      isDefault: vehicle.isDefault,
```
radoslavzeman, line 42: The PATCH request requires only `id` and changed attributes. This is the main difference between PATCH and PUT. Why we force to pass all attributes to this function if it is not required by the endpoint?

app/(app)/vehicles/edit-vehicle.tsx
radoslavzeman, line 1: The whole logic of this file is not very clean. The usual pattern in this case could be:
- check if search params exist (`vehicleId` in this case) and if it is in correct shape (`parseInt`)  - if not, return error screen
- check if the vehicle exists - if not, return error screen

Then you can be sure that id is in correct format and type (number) and that the vehicle exists in the rest of the component and you don't have to check it or parse it multiple times and return unintuitive voids or nulls.

app/(app)/vehicles/edit-vehicle.tsx
```diff
@@ -0,0 +1,86 @@
...
+            <TextInput autoCorrect={false} value={vehicleName} onChangeText={setVehicleName} />
+          </AccessibilityField>
+
+          <Button loading={isLoading} onPress={handleSaveVehicle}>
+            {t('editVehicle')}
+          </Button>
+
+          {!isPresented && (
+            <Link href="/" asChild>
+              <Button variant="plain-dark">Dismiss</Button>
```
radoslavzeman, line 77: We should put this into translation finally :D  (for each modal)

app/(app)/vehicles/edit-vehicle.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [vehicleName, setVehicleName] = useState(vehicle?.name ?? '')
+
+  // If the page was reloaded or navigated to directly, then the modal should be presented as
+  // a full screen page. You may need to change the UI to account for this.
+  const isPresented = router.canGoBack()
+
+  if (!vehicleId) return null
+
+  const handleSaveVehicle = async () => {
+    if (!vehicle) return
```
radoslavzeman, line 36: See the comment for the whole file.

state/VehiclesStoreProvider/VehiclesStoreProvider.tsx
```diff
@@ -12,9 +12,12 @@ export type AddVehicle = {
   isDefault?: boolean
 }
 
+export type EditVehicle = AddVehicle & { id: number }
```
radoslavzeman, line 15: Check the comment about PATCH request in `editVehicle` function.

state/VehiclesStoreProvider/VehiclesStoreProvider.tsx
```diff
@@ -74,6 +77,20 @@ const VehiclesStoreProvider = ({ children }: PropsWithChildren) => {
...
   )
 
+  const editVehicle = useCallback(
+    async (vehicle: EditVehicle) => {
+      await updateMutation.mutateAsync({
+        id: vehicle.id,
+        updateVehicleDto: {
+          vehiclePlateNumber: vehicle.licencePlate,
+          name: vehicle.vehicleName,
+          isDefault: !!vehicle.isDefault,
```
radoslavzeman, line 87: It will probably be changed according to PATCH request, but in general, we want to avoid unnecessary type "changes" as this. There is no need to use `!!` (cast it to boolean) (`vehicle.isDefault` can be `true`, `false` or `undefined` an it is completely okay because the same type is accepted by the endpoint).

radoslavzeman (approved):
There is few more comments to address this time :)

radoslavzeman (changes requested):
Accidentally approved :D

PavelFilo:
Screenshot 2024-04-09 at 16 04 22

 I've created an error screen and added vehicleId and vehicle screen check into one if check in order to have less redundant code (otherwise it has to be there twice). Also, I haven't used parseInt but Number(), because of the better strictness it offers. The last change was to remove redundant code with API edit request

components/controls/vehicles/EditVehicleForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+const EditVehicleForm = ({ vehicle }: Props) => {
+  const t = useTranslation('VehiclesScreen')
+  const { editVehicle, isLoading } = useVehiclesStoreContext()
+
+  const [vehicleName, setVehicleName] = useState(vehicle?.name ?? '')
+
+  const handleSaveVehicle = async () => {
+    if (!vehicle) return
```
radoslavzeman, line 27: Can `vehicle` be not defined?

components/controls/vehicles/EditVehicleForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+        {/* eslint-disable-next-line react/style-prop-object */}
+        <StatusBar style="light" />
+
+        <ScreenContent>
+          <AccessibilityField label={t('licencePlateFieldLabel')}>
+            <TextInput
+              autoCapitalize="characters"
+              autoCorrect={false}
+              isDisabled
+              value={vehicle?.vehiclePlateNumber}
```
radoslavzeman, line 49: Probably unnecessary `?` after moving to `EditVehicleForm` component? Please check the whole file :)

components/screen-layout/ErrorScreen.tsx
```diff
@@ -0,0 +1,32 @@
...
+} & Omit<ScreenViewProps, 'children'>
+
+/**
+ * Error screen with error message and title
+ */
+const ErrorScreen = ({ title, text, actionButton, ...rest }: Props) => {
+  const t = useTranslation('ErrorScreen')
+
+  return (
+    <ScreenViewCentered options={{ headerTransparent: true }} title="" {...rest}>
```
radoslavzeman, line 21: Do we still need to provide `title=""`? (Just asking, not mandatory to solve for this PR)

state/VehiclesStoreProvider/VehiclesStoreProvider.tsx
```diff
@@ -12,9 +12,12 @@ export type AddVehicle = {
   isDefault?: boolean
 }
 
+export type EditVehicle = { id: number; vehicleName?: string; isDefault?: boolean }
```
radoslavzeman, line 15: Keep formatting in multiple lines please :)

state/VehiclesStoreProvider/VehiclesStoreProvider.tsx
```diff
@@ -74,6 +77,19 @@ const VehiclesStoreProvider = ({ children }: PropsWithChildren) => {
     [insertMutation, vehicles],
   )
 
+  const editVehicle = useCallback(
+    async (vehicle: EditVehicle) => {
+      await updateMutation.mutateAsync({
+        id: vehicle.id,
+        updateVehicleDto: {
+          name: vehicle.vehicleName,
+          isDefault: vehicle.isDefault,
```
radoslavzeman, line 86: Editing `isDefault` is probably never used?

radoslavzeman (approved):
Nice and clean :)

PR #482, opened 2024-04-30
package upgrades

PavelFilo:
maintenance of packages


package.json
radoslavzeman, line 1: Please keep all `dependencies` on fixed version or at least on fixed minor version (using `~`, not `^`)

utils/paymentRedirect.ts
```diff
@@ -16,8 +16,8 @@ export const paymentRedirect = (ticketInit: TicketInitDto, paymentOption: Paymen
       paymentOption === 'apple-pay'
         ? ticketInit.paymentUrls.paymentUrlAPAY
         : paymentOption === 'google-pay'
-        ? ticketInit.paymentUrls.paymentUrlGPAY
-        : ticketInit.paymentUrls.paymentUrlCard
+          ? ticketInit.paymentUrls.paymentUrlGPAY
+          : ticketInit.paymentUrls.paymentUrlCard
```
radoslavzeman, line 20: Just checking formatting, I think this is not forced by prettier, but by you IDE and it should not be. (Not a blocker for merging)

radoslavzeman (approved):
🙌

utils/paymentRedirect.ts
PavelFilo: this was an error in ts:check

PR #498, opened 2024-05-22
Switch to new validate TS pipeline from Bratislava's action repo

vidriduch:


.github/workflows/validate.yml
```diff
@@ -0,0 +1,29 @@
...
+on: [ pull_request ]
+
+jobs:
+  validate:
+    name: Validate typescript
+    uses: bratislava/github-actions/.github/workflows/validate-typescript.yml@stable
+    with:
+      workdir: '.'
+
+  typescript-check:
```
radoslavzeman, line 12: `eslint-check`?

PR #539, opened 2024-06-12
Cleanups, fix typos

radoslavzeman:


hooks/useSignInOrSignUp.ts
```diff
@@ -65,7 +65,7 @@ export const useSignInOrSignUp = () => {
         /** Try to sign in the user. Cognito will throw an error for non-registered user. */
         await signInAndRedirectToConfirm(phone)
       } catch (error) {
-        console.log('signInAndRedirectToConfirm error', error, JSON.stringify(error))
+        // Expected error 'UserNotFoundException' means non-registered user
```
PavelFilo, line 68: do we need it twice as the same message is 3 lines below?

app/(auth)/sign-in/index.tsx
```diff
@@ -34,21 +34,21 @@ const Page = () => {
   const [selectedCountry, setSelectedCountry] = useUsedCountryStorage()
 
   useEffect(() => {
-    if (selectedCountry) return
-
-    setSelectedCountry('SK')
+    if (!selectedCountry) {
```
PavelFilo, line 37: why don't we use first approach? I think it's more readable when used consistently like that

PavelFilo (approved):
lgtm

PR #541, opened 2024-06-12
Rewrite notification permission code to simpler functions

radoslavzeman:
The aim of this rewrite is to keep the same functionality and behaviour, but write it in cleaner form (divide responsibilities into separate functions and hooks). Any functionality updates should be done in further PRs.

### DIFF
--- a/modules/map/hooks/useNotificationPermission.ts
+++ b/modules/map/hooks/useNotificationPermission.ts
@@ -1,49 +1,21 @@
 import messaging from '@react-native-firebase/messaging'
-import { useMutation, useQuery } from '@tanstack/react-query'
 import * as Device from 'expo-device'
-import { useCallback, useEffect, useState } from 'react'
-import { Platform } from 'react-native'
+import { useEffect, useState } from 'react'
 
-import { clientApi } from '@/modules/backend/client-api'
-import { devicesOptions } from '@/modules/backend/constants/queryOptions'
-import { MobileDevicePlatform } from '@/modules/backend/openapi-generated'
+import { useRegisterDevice } from '@/modules/map/hooks/useRegisterDevice'
 import { PermissionStatus } from '@/utils/types'
 
-type Options =
-  | {
-      autoAsk?: boolean
-      skipTokenQuery?: boolean
-    }
-  | undefined
+export const useNotificationPermission = (options?: { skipTokenRegistration?: boolean }) => {
+  const { skipTokenRegistration = false } = options ?? {}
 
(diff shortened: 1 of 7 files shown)


modules/map/hooks/useNotificationPermission.ts
```diff
@@ -1,99 +1,61 @@
...
-    const authStatus = await messaging().hasPermission()
-    const enabled =
-      authStatus === messaging.AuthorizationStatus.AUTHORIZED ||
-      authStatus === messaging.AuthorizationStatus.PROVISIONAL
+  const { registerDeviceIfNotExists } = useRegisterDevice(skipTokenRegistration)
 
-    if (enabled) {
-      setPermissionStatus(PermissionStatus.GRANTED)
+  useEffect(() => {
+    const checkStatus = async () => {
```
PavelFilo, line 18: aren't functions more readable when outside useEffect? what about adding it into useCallback and outside of useEffect for better readability

modules/map/hooks/useRegisterDevice.ts
```diff
@@ -0,0 +1,33 @@
...
+      }),
+    onSuccess: async () => {
+      await devicesQuery.refetch()
+    },
+  })
+
+  const registerDeviceIfNotExists = async () => {
+    const token = await messaging().getToken()
+
+    if (token && !devicesQuery.data?.devices.some((device) => device.token === token)) {
```
PavelFilo, line 27: if data are undefined the mutation triggers, do we want it to happen? so if the devicesQuery throws an error this mutation can create a loop we try to avoid.

modules/map/hooks/useNotificationPermission.ts
radoslavzeman: Are they?

modules/map/hooks/useRegisterDevice.ts
radoslavzeman: In javascript, `undefined` is "falsy" value, so it does not pass the condition and mutation is not triggered. We you this pattern heavilly in Next/Strapi projects.

See here, "yes" is not logged into browser console when `&& undefined` is used in condition:
image

But we could make it more readable and add  `&& devicesQuery.data` if you think so.

PavelFilo: Screenshot 2024-06-12 at 12 23 47
then you are doing it wrong :D

radoslavzeman: Fixed ✅

PavelFilo (approved):

Issue #527, opened 2024-06-06
Onboarding screen is not visible whole on small screens


radoslavzeman:
Test on iphone SE

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/e7f1b99b-3d53-46fc-a902-ae7710581ebf)

PR #552, opened 2024-06-17
#527: Fix onboarding hidden text on small screens

radoslavzeman:
Images keep the square shape if possible (if there is enough vertical space) (A),
or they shrink if there is not enough space (B).
(Borders are used just on screenshots, not in code)

A
image

B
image


PavelFilo (approved):
lgtm

PR #596, opened 2024-07-22
#591 Remove leading zero in login value

### LINKED ISSUE #591: Bug: User can log in with number in format `+42109...`
When user enters the number with leading 0, e.g. "0901123123", it's sent to Cognito as "+4210901123123" - containing a "0" character between country prefix and "901...".
The verification sms is sent to correct number (without 0), but user is signed in as a completely different user (with 0 and with different userId).

We should handle this situation by skipping the zero.

radoslavzeman:
- closes #591


PavelFilo (approved):
I don't like it but also I don't have any other solution

PR #597, opened 2024-07-22
Private/#5: Fix notification settings after permissions are allowed

radoslavzeman:
- closes bratislava/paas-mpa-private#5


PavelFilo (approved):
isn't true default for this?

PR #604, opened 2024-07-24
#531 Temporary fix for BottomSheet to show also when ReducedMotion is on

### LINKED ISSUE #531: Android user: nejde kliknut na 3 bodky pri edite vozidla

radoslavzeman:
- fix `BottomSheet` didn't appear on three-dots button press in Vehicles screen and Ticket History screen
- closes #531 
- closes bratislava/paas-mpa-private#1

The issue is probably caused when ReducedMotion is enabled. Note that `ReducedMotion` is also enabled when battery is <10% (on ios).

Quick explanation of the problem and the fix:
With newer version of `react-native-reanimated` that included support for `ReducedMotion`, `BottomSheet` has problem to appear if `ReducedMotion` is enabled. It's supposed to be fixed in BottomSheet version 4.6.3 (latests at this time), but [people write](https://github.com/gorhom/react-native-bottom-sheet/issues/1560#issuecomment-2183549675) it's still buggy, and the upgrade to newer version should be done with proper testing, not as quick fix.
So I decided to use [the hot-fix](https://github.com/gorhom/react-native-bottom-sheet/issues/1560#issuecomment-1750466864): disable `animateOnMount` when `ReducedMotion` is used.
People also mentioned problem with this hot-fix with `.dismiss()`, but we use `.close()` instead of `.dismiss()` and it worked okay when I tested it.

Tested locally on iphone 15 on physical device with ReducedMotion enabled ✅ 

See more in the issue thread: https://github.com/gorhom/react-native-bottom-sheet/issues/1560

### DIFF
--- a/app/(app)/payment-options.tsx
+++ b/app/(app)/payment-options.tsx
@@ -5,6 +5,7 @@ import {
 } from '@gorhom/bottom-sheet'
 import { useCallback, useRef, useState } from 'react'
 import { Platform, ScrollView } from 'react-native'
+import { useReducedMotion } from 'react-native-reanimated'
 
 import PaymentOptionRow from '@/components/controls/payment-methods/rows/PaymentOptionRow'
 import { PaymentOption } from '@/components/controls/payment-methods/types'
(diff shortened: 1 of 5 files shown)


PavelFilo (approved):
in future we should move BottomSheetModal with default values to separate component

Issue #594, opened 2024-07-17
Android: Handle push notifications properly


zuzanaslobodova:
https://github.com/invertase/react-native-firebase/issues/2031#issuecomment-1403048986

PR #611, opened 2024-07-26
#594 Request notification permission on Android

radoslavzeman:
- minor renaming in first commit
- add missing request for notification permission on Android

### DIFF
--- a/modules/map/hooks/useNotificationPermission.ts
+++ b/modules/map/hooks/useNotificationPermission.ts
@@ -2,12 +2,15 @@ import messaging from '@react-native-firebase/messaging'
 import { useMutation } from '@tanstack/react-query'
 import * as Device from 'expo-device'
 import { useCallback, useEffect, useState } from 'react'
+import { Platform } from 'react-native'
 
 import { clientApi } from '@/modules/backend/client-api'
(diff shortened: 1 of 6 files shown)


modules/map/hooks/useNotificationPermission.ts
```diff
@@ -42,29 +45,49 @@ export const useNotificationPermission = () => {
...
+        setPermissionStatus(UnifiedPermissionStatus.DENIED)
+      }
+    }
+
+    // https://rnfirebase.io/messaging/usage#android---requesting-permissions
+    if (Platform.OS === 'android') {
+      await requestNotificationPermissionAndroidAsync()
+
+      // Android doesn't return value of current permission status as messaging().requestPermission() does, so we need to check it manually
+      const authStatus = await messaging().hasPermission()
```
radoslavzeman, line 76: This is the main difference from PR #594 where we incorrectly compared the result of `requestNotificationPermissionAndroidAsync` - this does not return actual PermissionStatus, it returns the status of native modal request.
Therefore, we have to call `messaging().hasPermission()` to get the actual PermissionStatus.

The whole hook needs refactoring together with Permissions Screen, so a little bit of duplicate code here should not be a problem.

PavelFilo (approved):
Lgtm

Issue #612, opened 2024-07-26
Fix Bottom Sheets when `ReducedMotion` is enabled


radoslavzeman:
Bottom Sheets do not show up when `ReducedMotion` setting is enabled (verified on ios, but there is probably similar issue on android).
Note that ReducedMotion is also enabled when battery is <10% (on ios).

This caused problems manetioned in #531 and bratislava/paas-mpa-private#1.

Quick explanation of the problem and the fix:
With newer version of `react-native-reanimated` that included support for `ReducedMotion`, `BottomSheet` has problem to appear if ReducedMotion is enabled. It's supposed to be fixed in BottomSheet version 4.6.3 (latests at this time), but https://github.com/gorhom/react-native-bottom-sheet/issues/1560#issuecomment-2183549675 it's still buggy, and the upgrade to newer version should be done with proper testing, not as quick fix.

See more in the issue thread: https://github.com/gorhom/react-native-bottom-sheet/issues/1560

We fixed some cases where `BottomSheetModal` is used in #604
> So I decided to use https://github.com/gorhom/react-native-bottom-sheet/issues/1560#issuecomment-1750466864: disable animateOnMount when ReducedMotion is used.
> People also mentioned problem with this hot-fix with .dismiss(), but we use .close() instead of .dismiss() and it worked okay when I tested it.

This issue also happens with `BottomSheet` (not the modal version), that still has to be fixed.

PR #613, opened 2024-07-31
#612 Fix BottomSHeets when Reduced Motion enabled

radoslavzeman:
- add the same hotfix to all `BottomSheet`s (before it was added only to `BottomSheetModal`s) as in #604 
- move `key` to fragment in `timeframes` bottom sheet, because it showed error "duplicate key" 🤷


PavelFilo (approved):
Looks nice, and please create task for the filters refactor, because we will probably forget about that 😄 it's kind of weird that it is screen and returns just bottom sheet, it shouldn't be like that...

PR #634, opened 2024-08-27
#633 Show phone number in confirm sign-in screen text

### LINKED ISSUE #633: Show entered phone number on confirm sign-in screen
Show entered phone number on confirm sign-in screen so user can double-check if they entered the correct number while waiting for the code.

radoslavzeman:
- closes #633 

image


PavelFilo (commented):
good job :)

PR #643, opened 2024-09-04
Use sanitized version of licence plate in one time licence plate

### LINKED ISSUE #638: FE: BPK + NPK upper case and trim EČV
User pridal BPK podla metabase do appky 23/8 o 21:52, listok pre EČV, na ktore je BPK vydaná aktivoval o 21:54. Kredit mal k dispozícii ešte celé 2H, zľava sa neuplatnila

https://github.com/user-attachments/assets/cf3c8a22-9f5b-4a2e-a432-3e68da9f233d

### LINKED ISSUE #641: Validate ECV in one-time use input
The one-time use ECV input does not validate the inserted ECV the same way as we do when new vehicle is added.

We should use the same validation in both cases (uppercase, trim).

PavelFilo:
closes #641 
closes #638


radoslavzeman (approved):
🙌

Issue #650, opened 2024-09-16
My location button disapears on rerender


PavelFilo:
bottom sheet top attachment position is not determined correctly after reanimated update

PR #651, opened 2024-09-16
#650 resolve issue with top attachment that disapears on rerender

PavelFilo:
- removed todo text from gothom bottom sheet library issue because it's fixed in the latest version, but now it's feature, not a hotfix to a bug 😄,
- the height state of BottomSheetTopAttachment in new version of react-native-reanimated starts to be 0 in hook useAnimatedStyle. In my opinion it's because it runs on UI thread and state on JS thread so the value differs


radoslavzeman (approved):
💯

PR #658, opened 2024-09-17
#370 add social into ParkingCard component

PavelFilo:
- run generate clients and
- add social type from API enum to ParkingCard component


radoslavzeman (approved):
👍

PR #665, opened 2024-09-27
#656 Add error monitoring and logging

PavelFilo:
Added Sentry as a frontend monitoring tool (Also added secret to Expo which is stated only in documentation of Sentry and not in our code and is added during build)


radoslavzeman (approved):
Nice :)

PR #675, opened 2024-10-08
#625 add pointer events to bottom sheet wrapper

PavelFilo:


radoslavzeman (approved):
Add a comment why is it needed

PR #676, opened 2024-10-09
#662 Add Apple live activities 🥳 

PavelFilo:
All hyped feature for iOS is finally here :)
the new code includes:
- native widgets (locked screen widget, dynamic island in closed, opened and minified state) 
- swift module that opens and updates widget with functions that are exported to JavaScript (TypeScript)
- functionality in React to open and close live activity

The Live Activity is opened on purchase and stored into local storage (it can also be added to BE eventually as field in ticketDB, but for now localStorage is fine)
The Live Activity updates if it exists and ticket is prolonged
The Live Activity is dismissed on shorten the ticket

In other states it just becomes stale as normal notification and user should remove it by himself as normal notification

This approach will be changed in the future when firebase adds ActivityKit functionality and we will eventually call it from BE.

PR also includes useEffectOnce which is nice feature to not disable ESLint rule and write unnecessary comments why it was done... now everyone can know that it should just trigger once it renders for the first time


components/purchase/PaymentStatus/SuccessContent.tsx
```diff
@@ -22,6 +26,12 @@ const SuccessContent = ({ ticket, purchaseType }: Props) => {
     },
   }
 
+  useEffectOnce(() => {
+    if (purchaseType === 'payment') {
+      startLiveActivity(ticket)
+    } else updateLiveActivity(ticket)
```
radoslavzeman, line 32: Consistent usage of braces:
```ts
} else {
  updateLiveActivity(ticket)
}
```

package.json
```diff
@@ -53,6 +54,7 @@
     "expo-linking": "~6.3.1",
     "expo-localization": "~15.0.3",
     "expo-location": "~17.0.1",
+    "expo-modules-core": "^1.12.24",
```
radoslavzeman, line 57: use `~` ?

package.json
```diff
@@ -20,6 +20,7 @@
   },
   "dependencies": {
     "@aws-amplify/react-native": "~1.0.6",
+    "@bacons/apple-targets": "^0.0.5",
```
radoslavzeman, line 23: use `~` ?

radoslavzeman (approved):
So happy to see this! ❤️ 

I think it could be very useful to add some basic info about live activities into readme. E.g. some liks to docs, what is in the repo etc.

PR #679, opened 2024-10-14
#424 specify redirect to the app from notifications

PavelFilo:
- add notification Handler component that checks when opening notifications from background state or from quit state
- split the tickets page into Lists and tabs that include lists
- refactor functionality of tabs to not use the state as an index, but router param (tab)


radoslavzeman (approved):
💯 nicee

Issue #212, opened 2023-11-30
Rate the app


radoslavzeman:
Rating should be done by "native" store components, not ours.

https://www.figma.com/file/3TppNabuUdnCChkHG9Vft7/paas-mpa?node-id=2631%3A12608&mode=dev

![Image](https://github.com/bratislava/paas-mpa/assets/19418224/aa212b02-fa12-4815-a0b0-f22c578716df)

PR #682, opened 2024-10-16
#212 Add rate app action row to about page

PavelFilo:
- refactor the about items to not support Link component, but router.navigate instead because some items need to not have Link there,
- add navigateToStore function that navigates to https links
- add navigateToStoreReviews function that has deeplink to stores which open reviews

### DIFF
--- a/app/(app)/about/index.tsx
+++ b/app/(app)/about/index.tsx
@@ -15,7 +15,7 @@ import { navigateToStoreReview } from '@/utils/navigateToStore'
 const navigateToRoute = (url: string) => {
   router.navigate({
     pathname: '/about/webview',
-    params: { webviewUri: url } as WebviewSearchParams,
+    params: { webviewUri: url } satisfies WebviewSearchParams,
   })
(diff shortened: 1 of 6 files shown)


app/(app)/about/index.tsx
```diff
@@ -10,56 +10,57 @@ import ScreenView from '@/components/screen-layout/ScreenView'
 import Divider from '@/components/shared/Divider'
 import PressableStyled from '@/components/shared/PressableStyled'
 import { useTranslation } from '@/hooks/useTranslation'
+import { navigateToStoreReview } from '@/utils/navigateToStore'
+
+const navigateToRoute = (url: string) => {
+  router.navigate({
+    pathname: '/about/webview',
+    params: { webviewUri: url } as WebviewSearchParams,
```
radoslavzeman, line 18: try to use `satisfies`, check other usages in this case

radoslavzeman (approved):
🐀 ⭐

Issue #377, opened 2024-01-18
Header has slight margin after removing one of the buttons


PavelFilo:
you don't have any vehicles and afterward, you add one. the title is a little bit uncentred

PR #683, opened 2024-10-17
#377 add empty icons to header when there is no icon

PavelFilo:
Placeholder removes the title typography's stretch when Icons are mounted or unmounted


radoslavzeman (approved):
Interesting

PR #703, opened 2024-11-29
#702 add invisible cloudflare captcha to login

PavelFilo:
Add the invisible captcha that's requested when the user presses the sign-in button and starts the login process.

The whole captcha is a web thing, so here we have it in Webview.
For now, it should be optional and not required in cognito, because we don't know with certainty, that it won't throw an error to regular users, so more research is needed.


MarekBodingerBA:
So we display (or don't display) the captcha but don't send the token anywhere, even if it fails or not?

If you just want to have the feature in the code but don't use it, I strongly suggest using [feature toggles](https://martinfowler.com/articles/feature-toggles.html).

.env
```diff
@@ -6,3 +6,4 @@ EXPO_PUBLIC_API_URL=https://paas-mpa-backend.staging.bratislava.sk
 EXPO_PUBLIC_MINIO_BUCKET=paas-mpa-dev
 EXPO_PUBLIC_DEPLOYMENT=development
 EXPO_PUBLIC_SENTRY_DNS=https://b58eb0384ab7010886ff5cae0417cad5@o701870.ingest.us.sentry.io/4508025288065024
+EXPO_PUBLIC_CAPTCHA_SITE_KEY=0x4AAAAAAAGcjNLteOzO0fDF
```
MarekBodingerBA, line 9: Imho, it's better to have envs like `PLAUSIBLE_KEY` vs `ANALYTICS_KEY`, so I would suggest `TURNSTILE_SITE_KEY` instead of `CAPTCHA_SITE_KEY`

components/inputs/Captcha.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { environment } from '@/environment'
+
+type Props = {
+  onSuccess: (token: string) => void
+  onFail: () => void
+}
+
+const Captcha = ({ onSuccess, onFail }: Props) => {
+  const handleMessage = (event: { nativeEvent: { data: string } }) => {
```
MarekBodingerBA, line 11: Is there a reason not to use `WebViewNativeEvent` from `react-native-webview` typings?

components/inputs/Captcha.tsx
```diff
@@ -0,0 +1,59 @@
...
+      source={{
+        baseUrl: 'https://bratislava.sk',
+        html: `
+            <!DOCTYPE html>
+            <html>
+                <head>
+                    <script src="https://challenges.cloudflare.com/turnstile/v0/api.js?onload=_renderCaptcha" async defer></script>
+                </head>
+                <body>
+                    <div>aaaa</div>
```
MarekBodingerBA, line 34: Probably redundant.

components/inputs/Captcha.tsx
```diff
@@ -0,0 +1,59 @@
...
+                    <div id="turnstileWidget"></div>
+                    <div>aaa</div>
```
MarekBodingerBA, line 36: Probably redundant.

MarekBodingerBA:
I think you need to think about how to display the error to the user, I don't think it's possible to skip error handling, two quick ideas:
1. Convert the whole screen to WebView to display turnstile as is. (not preferable probably)
2. Try to detect when it fails an appropriately display (visually) web view (probably would needs some resizing and testing) so user can handle the case. 

There's an example failure token that you can use for testing - https://developers.cloudflare.com/turnstile/troubleshooting/testing/.

components/inputs/Captcha.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * Captcha component handles Cloudflare's invisible captcha. The webview is opened without any user interaction. The user is not aware of the captcha, but the token is sent to the onSuccess callback.
+ * TODO: For now the captcha is not required by cognito, we need the number of users who failed the captcha and why. After that the errors will be handled and the captcha can be enabled in cognito.
+ * @param onSuccess Callback function that is called when the captcha is successfully solved. The token is passed as an argument.
+ * @param onFail Callback function that is called when the captcha fails. The error code is passed as an argument and sent to the cognito.
+ */
+const Captcha = ({ onSuccess, onFail }: Props) => {
+  const handleMessage = (event: WebViewMessageEvent) => {
+    const { data } = event.nativeEvent
+
+    if (data.split('_')[0] === 'error') {
```
MarekBodingerBA, line 20: I would just use `data.startsWith("error_")`

MarekBodingerBA (approved):

MarekBodingerBA (approved):

PR #708, opened 2024-12-09
#702 Captcha second stage integration

PavelFilo:
The second stage includes two tries and then the signup.
We found some of the captcha verifications fail (10 in 5 days) so we want to know whether they will fail on second try, because turnstile docs suggest to retry the captcha on errors like those.

PR includes
- add retry to captcha after one failed one,
- add error message so user knows what is happening

### DIFF
--- a/app/(auth)/sign-in/index.tsx
+++ b/app/(auth)/sign-in/index.tsx
@@ -1,4 +1,3 @@
-import * as Sentry from '@sentry/react-native'
 import { useEffect, useRef, useState } from 'react'
 import { ScrollView, View } from 'react-native'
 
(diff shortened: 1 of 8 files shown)


MarekBodingerBA:
It's a bit hard follow captcha logic (retry, rerender and so on) and I think it shouldn't be leaking to the sign in page. You can add `wasRetried` flag and act upon it. Also using the `key` to trigger a rerender is common practise that you can use.

Something like this: https://chatkit.app/s/kgguym0iupue800

MarekBodingerBA (changes requested):
.

app/(auth)/sign-in/index.tsx
radoslavzeman, line 1: Please doublecheck if change from `phoneWithoutSpaces` to `phone` is 100% correct. I see the issue issue and TODO comment with `phoneWithoutSpaces`. But `phone` can contain spaces - please double-check, if this will not cause problems.

radoslavzeman (approved):
Tell me if you can merge with this approval or not :)

PR #721, opened 2024-12-18
#719 add isWinterHolidayPeriod check to map feature

PavelFilo:
remove TODO flag because we can keep the code the whole year


radoslavzeman (approved):
If we want to keep it there the whole year, maybe we could call it "Christmas market" or "Christmas market tree" or "Christmass tree" instead of just "tree". I'm sure "tree" will be confusing during the year, when there is no christmas vibe :D

PR #728, opened 2025-01-29
#727 remove priceObject useMemo and cleanup useEffects

PavelFilo:
- remove price object useMemo,
- move refetch query interval inside queryOptions instead of useEffect,
- remove another redundant useEffect and rename vehicle state (i couldn't help myself and did it here)

### DIFF
--- a/app/(app)/(purchase-and-payment)/purchase/index.tsx
+++ b/app/(app)/(purchase-and-payment)/purchase/index.tsx
@@ -1,6 +1,6 @@
 import { useMutation, useQuery } from '@tanstack/react-query'
-import { Link, router, useFocusEffect } from 'expo-router'
-import { useCallback, useEffect, useMemo, useState } from 'react'
+import { Link, router } from 'expo-router'
+import React, { useEffect, useState } from 'react'
 import { ScrollView } from 'react-native'
 import { useDebounce } from 'use-debounce'
 
@@ -28,10 +28,7 @@ import {
   ticketPriceOptions,
   verifiedEmailsLengthOptions,
 } from '@/modules/backend/constants/queryOptions'
-import {
-  GetTicketPriceRequestDto,
-  InitiatePaymentRequestDto,
-} from '@/modules/backend/openapi-generated'
+import { InitiatePaymentRequestDto } from '@/modules/backend/openapi-generated'
 import { usePurchaseStoreContext } from '@/state/PurchaseStoreProvider/usePurchaseStoreContext'
 import { usePurchaseStoreUpdateContext } from '@/state/PurchaseStoreProvider/usePurchaseStoreUpdateContext'
 import { useVehiclesStoreContext } from '@/state/VehiclesStoreProvider/useVehiclesStoreContext'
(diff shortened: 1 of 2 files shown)


app/(app)/(purchase-and-payment)/purchase/index.tsx
```diff
@@ -138,17 +115,15 @@ const PurchaseScreen = () => {
...
-      return
-    }
+    if (!licencePlate) return
 
     const actualPaymentOption = paymentOption ?? defaultPaymentOption
 
     initPaymentMutation.mutate(
       {
-        ...priceRequestBody,
+        ...createPriceRequestBody({ udr, licencePlate, duration: debouncedDuration, npk, locale }),
```
radoslavzeman, line 126: Please send the same createPriceRequestBody

app/(app)/(purchase-and-payment)/purchase/index.tsx
```diff
@@ -45,7 +42,7 @@ const PurchaseScreen = () => {
   // TODO: find solution for height of bottom content with drawing
   const [purchaseButtonContainerHeight, setPurchaseButtonContainerHeight] = useState(0)
 
-  const [hasLicencePlateError, setHasLicencePlateError] = useState(false)
+  const [isTouched, setIsTouched] = useState(false)
```
radoslavzeman, line 45: What is touched? :D Please use more specific name.

modules/backend/constants/queryOptions.ts
```diff
@@ -115,27 +113,24 @@ export const verifiedEmailsInfiniteOptions = (options?: PageSize) => {
...
+      body.ticket.udrUuid,
+      body.ticket.parkingStart,
+    ],
     queryFn: () => clientApi.ticketsControllerGetTicketPrice(body),
     select: (res) => res.data,
     // https://tanstack.com/query/latest/docs/react/guides/migrating-to-v5#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function
     placeholderData: keepPreviousData,
-    enabled: !!udr && !!licencePlate && !!duration,
+    refetchInterval: 1000 * 60,
+    refetchIntervalInBackground: true,
```
radoslavzeman, line 132: Do we really want to refetch it in background? Isn't use `useQueryWithFocusRefetch` enough? Can refetching in background cause significant increase in number of requests?

modules/backend/constants/queryOptions.ts
```diff
@@ -115,27 +113,24 @@ export const verifiedEmailsInfiniteOptions = (options?: PageSize) => {
...
+      body.npkId,
+      body.ticket.udrUuid,
+      body.ticket.parkingStart,
+    ],
     queryFn: () => clientApi.ticketsControllerGetTicketPrice(body),
     select: (res) => res.data,
     // https://tanstack.com/query/latest/docs/react/guides/migrating-to-v5#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function
     placeholderData: keepPreviousData,
-    enabled: !!udr && !!licencePlate && !!duration,
+    refetchInterval: 1000 * 60,
```
radoslavzeman, line 131: Hmm it's interesting we didn't do it already :D

app/(app)/(purchase-and-payment)/purchase/index.tsx
PavelFilo: the whole form is touched (or was submitted in this case), this naming is common within forms and we don't have any other form in there :) I can change it to wasSubmitted/isSubmitted if you find it more clear.

radoslavzeman: Okay, thanks for the explanation :) `isSubmitted` seems to be more clear :)

radoslavzeman (approved):
🙌 💪

PR #730, opened 2025-02-21
#706 change day difference to include last day

PavelFilo:
- change formatDate function to formatValidUntilDate function which handles the case of day difference

### DIFF
--- a/translations/sk.json
+++ b/translations/sk.json
@@ -158,7 +158,7 @@
   "ParkingCards.type.subscriber": "Abonentská karta",
   "ParkingCards.type.tzp": "ŤZP karta",
   "ParkingCards.type.visitor": "Návštevnícka karta",
-  "ParkingCards.validUntil": "Platnosť do",
+  "ParkingCards.validUntil": "Platnosť do (vrátane)",
   "ParkingCardTypeRow.type.all": "Všetky parkovacie karty",
   "ParkingCardTypeRow.type.bonusCards": "Iba bonusové karty",
   "ParkingCardTypeRow.type.visitorCards": "Iba návštevnícke karty",
--- a/translations/sk.json
+++ b/translations/sk.json
@@ -158,7 +158,7 @@
   "ParkingCards.type.subscriber": "Abonentská karta",
   "ParkingCards.type.tzp": "ŤZP karta",
   "ParkingCards.type.visitor": "Návštevnícka karta",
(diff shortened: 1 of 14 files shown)


translations/sk.json
```diff
@@ -158,7 +158,7 @@
   "ParkingCards.type.subscriber": "Abonentská karta",
   "ParkingCards.type.tzp": "ŤZP karta",
   "ParkingCards.type.visitor": "Návštevnícka karta",
-  "ParkingCards.validUntil": "Platnosť do",
+  "ParkingCards.validUntil": "Platnosť do (vrátane)",
```
radoslavzeman, line 161: I'm against this wording. It's redundant and we should not introduce it anywhere. As we discussed, the date is always considered "vratane" in SK laws.

utils/formatDate.ts
radoslavzeman, line 1: Isn't it useful to keep this basic formatting function even if it's not used?

utils/formatValidUntilDate.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Formats the validUntil date to the last included date in the format dd.MM.yyyy
+ * @param validUntil example: 2025-12-30T22:00:00.000Z
+ * @returns 30.12.2022
+ */
+export const formatValidUntilDate = (validUntil: string, locale: string) => {
+  // We subtract one minute because after localization the validUntil is 00:00 and we show last included date
```
radoslavzeman, line 7: From this comments, I'm not I understand the problem and the solution.

utils/formatValidUntilDate.ts
```diff
@@ -0,0 +1,15 @@
...
+  const time = new Date(validUntil).getTime() - 1 * 1000
+
+  return new Date(time).toLocaleDateString(locale, {
+    day: 'numeric',
+    month: 'numeric',
+    year: 'numeric',
+  })
```
radoslavzeman, line 14: I'm thinking if we shouldn't show also 23.59. Not because I think it needed in general, but bucause it's shown in Parkdots, so people are used to it.

radoslavzeman (approved):

utils/formatDate.ts
PavelFilo: i think that it was created because of the cards. And i don't see any point keeping it there, it's easy to redo. And also it's in the github history.

PR #733, opened 2025-03-07
#732 Change captcha from invisible to managed 

PavelFilo:
- change captcha to managed,
- remove retry of captcha and don't send user to cognito if captcha fails,
- add prefix to signIn function so we can work with this in lambda,
- change the function parameters to one object of values,
- version change to 1.7.0,
- change env for staging and dev turnstile site key


radoslavzeman (approved):
Looks promising! 🤝

PR #761, opened 2025-05-13
#757 Upgrade react-native to new architecture

PavelFilo:
### Expo Router changes (react-navigation)
- remove `navigation.reset()` in places where it can be removed, 
- adjust behaviour change of `router.navigate()` so the app works the same (navigate function no longer goes back if there is route in current stack... now the `dismissTo()` function handles this responsibility),
- change prop management in `react-native-tab-view` which uses `react-navigation` (similarly to `expo-router`) under the hood
### Map adjustments
- swap flyTo function for the one from library,
- remove `localFlyToCenter` and `flyToCenter` states,
- add `bbox` boundaries to geolocation api call,
### Splash screen
- remove full screen splash screen because android will not support it any longer and use only logo,
### Adjust tests
- change the zone to which the `purchase` and `prolong` tests are pointing


app/(app)/(purchase-and-payment)/purchase/payment.tsx
```diff
@@ -69,9 +69,9 @@ const PaymentScreen = () => {
       show('Unable to open payment URL.', { variant: 'danger' })
     }
 
-    redirectToPurchaseResult()
```
radoslavzeman, line 72: If the function is not used anymore, don't forget to remove it.

modules/map/utils/reverseGeocode.ts
```diff
@@ -3,13 +3,17 @@ import { Position } from 'geojson'
 
 import { environment } from '@/environment'
 import { GeocodingFeature } from '@/modules/arcgis/types'
+import { CITY_BOUNDS } from '@/modules/map/constants'
 
 export const reverseGeocode = async (position: Position) => {
   const [lon, lat] = position
   const accessToken = environment.mapboxPublicKey
   const limit = 1
```
radoslavzeman, line 11: remove? :)

radoslavzeman (commented):
So far very nice rewrite 👐

PR #767, opened 2025-06-24
#749 Add questioner page 

PavelFilo:
- regenerate api types,
- add questioner context so the query is fetched only once (for menu and modal),
- add questioner modal which shows text from database about questioner,
- add functionality to show questioner modal only once and flag it with shown keyword in local mobile storage,
- add page with webview where questioner externalUrl is shown,

the questioner is added via post request in swagger and azure identification.


app/(app)/questionnaire/[id].tsx
```diff
@@ -6,36 +6,36 @@ import { WebView } from 'react-native-webview'
...
 
-  if (isError || !isIdValid) {
-    return <ErrorScreen text={isError ? t('questioner.error.text') : t('questioner.error.empty')} />
+  if (isError) {
+    return <ErrorScreen text={t('QuestionnaireScreen.error.text')} />
+  }
+
+  if (!isIdValid) {
+    return <ErrorScreen text={t('QuestionnaireScreen.error.empty')} />
   }
```
Comment (radoslavzeman) [inline app/(app)/questionnaire/[id].tsx:39]: I would just move this above `if (isPending)`

app/(app)/menu.tsx
```diff
@@ -71,15 +71,11 @@ const MainMenuScreen = () => {
...
-          {
-            label: questioner.title,
-            icon: 'assistant',
-            path: `/questioner/${questioner.id}`,
-          } as const,
-        ]
-      : []),
+    ...(questionnaires?.map((questionnaire) => ({
+      label: questionnaire.title,
+      icon: 'assistant' as IconName,
```
radoslavzeman, line 76: `'assistant' as const` would work?

Issue #768, opened 2025-07-08
Change android targetSdkVersion to keep up with Google Play policies


PavelFilo:
The Google play requires for all the new updates of application to have minimal targetSdkVersion of 35 (android 15)

PR #770, opened 2025-07-08
#768 upgrade expo to upgrade targetSdkVersion

PavelFilo:
All android application updates needs to be on targetSdkVersion 35 which is default with expo sdk 53, that's why there is another update


package.json
```diff
@@ -19,77 +19,76 @@
...
+    "expo-router": "~5.1.2",
+    "expo-secure-store": "~14.2.3",
+    "expo-status-bar": "~2.2.3",
+    "expo-updates": "~0.28.16",
     "geojson": "~0.5.0",
     "i18next": "~23.11.2",
     "intl-pluralrules": "~2.0.1",
     "nativewind": "^4.0.36",
-    "react": "18.3.1",
+    "react": "19.0.0",
```
radoslavzeman, line 68: Why not `19.1.0`? :)

package.json
```diff
@@ -19,77 +19,76 @@
...
     "@gorhom/bottom-sheet": "~5.1.2",
     "@gorhom/portal": "~1.0.14",
     "@js-joda/core": "~5.5.3",
-    "@react-native-async-storage/async-storage": "1.23.1",
+    "@react-native-async-storage/async-storage": "2.1.2",
     "@react-native-community/netinfo": "11.4.1",
-    "@react-native-firebase/app": "^21.14.0",
-    "@react-native-firebase/messaging": "^21.14.0",
+    "@react-native-firebase/app": "^22.2.1",
+    "@react-native-firebase/messaging": "^22.2.1",
```
radoslavzeman, line 32: Just noticed we use `^` for these two packages, maybe switch to `~`.

radoslavzeman (approved):
Nice upgrade! Thank! 🫶

PR #771, opened 2025-07-08
#769 Add initialRouteName to fix deep links

PavelFilo:
after going to application with deep link (to /tickets) we don't provide back button to return to index screen because router will put it as only item in stack, therefore we need to add initialRouteName to enable this in app layout

There was other option with useEffect and setting index route with each deep link (with `navigation.reset(...)`), but it's not good option.


radoslavzeman (approved):
Nice!
Please provide a link to docs into code comment if possible.

PR #782, opened 2025-08-25
Fix parking cards

PavelFilo:
- remove unnecessary conditions from tests (now that BE is fixed the paygate is in language of user),
- remove wrong condition from parking cards... we don't need the second pard (it was wrong),
- show different result screen based on number of cards added

### DIFF
--- a/app/(app)/parking-cards/verification/index.tsx
+++ b/app/(app)/parking-cards/verification/index.tsx
@@ -139,12 +139,7 @@ const Page = () => {
 
                 return (
                   <PressableStyled key={panel} onPress={() => handlePanelPress(cardTypeInner)}>
-                    <ParkingCardTypeRow
-                      variant={panel}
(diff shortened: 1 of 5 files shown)


app/(app)/parking-cards/verification/index.tsx
```diff
@@ -139,12 +139,7 @@ const Page = () => {
 
                 return (
                   <PressableStyled key={panel} onPress={() => handlePanelPress(cardTypeInner)}>
-                    <ParkingCardTypeRow
-                      variant={panel}
-                      selected={
-                        cardType === cardTypeInner || (panel === 'all' && cardTypeInner === null)
-                      }
-                    />
+                    <ParkingCardTypeRow variant={panel} selected={cardType === cardTypeInner} />
```
shegood, line 142: Why did you change this logic?

PR #786, opened 2025-11-05
Escape slashes in english time format

shegood:


components/tickets/PurchaseBottomContent.tsx
```diff
@@ -75,9 +76,13 @@ const PurchaseBottomContent = ({
...
-              {t('PurchaseScreen.warnings.differentEnd', {
-                time: formatTime(new Date(priceQuery.data.ticketEnd), locale),
-              })}
+              <Trans
+                t={t}
+                tOptions={{ time: formatTime(new Date(priceQuery.data.ticketEnd), locale) }}
+                shouldUnescape
+              >
+                PurchaseScreen.warnings.differentEnd
+              </Trans>
```
PavelFilo: You are right, we don't solve anything with using `<Trans` component instead of function. The usage in JSON is small so person after us will need effort to find out what it means (`{{-time}}`). We can use it like this:
```suggestion
              {t('PurchaseScreen.warnings.differentEnd', {
                time: formatTime(new Date(priceQuery.data.ticketEnd), locale),
                interpolation: {
                  escapeValue: false,
                },
              })}
```

PavelFilo (changes requested):

PR #833, opened 2026-04-21
Prevent unintended user logouts

mpinter:
What was done to make the unintended logouts better:
- avoid calling `getCurrentAuthenticatedUser` (and underlying `getCurrentUser`) when not connected to internet - this _may_ cause clearing of valid refresh tokens from store - https://github.com/aws-amplify/amplify-js/issues/10393. But even more straightforward issue was that it caused navigation to login screen when the internet was not available on app startup, causing users to relogin even if it reconnected shortly
- updating amplify - likely won't have an effect as we were already on aws-amplify-6.15.8 (because of ~ version, see yarn.lock)
- changing refresh token lifespan to 6 years (from 180 days) - outside of this PR, in aws console

In addition this pins the correct version of yarn in volta and corepack, without which the package installations fail.

Where I need review the most:
- I'm only avoiding `getCurrentAuthenticatedUser` in the couple places this may be an issue, instead of preventing it from calling the problematic aws-amplify function from inside. Opens this up to problems in the future, but I don't want to muddle the functionality of the `getCurrentAuthenticatedUser` itself. But happy to discuss
- the hook in AuthProvider changed from simple 'useEffectOnce' to 'runThisOnceThisConditionIsMet' - let me know if I'm doing it in a react idiomatic way :)
- anything I'm missing

Only tested on iOs simulator, will need real device testing before release.

### DIFF
--- a/utils/languageDetectorPlugin.ts
+++ b/utils/languageDetectorPlugin.ts
@@ -1,3 +1,5 @@
+import NetInfo from '@react-native-community/netinfo'
+
 import { getCurrentAuthenticatedUser } from '@/modules/cognito/utils'
 import { getAndChangeUserLanguage } from '@/utils/getAndChangeUserLanguage'
 import { getAppLocale } from '@/utils/getAppLocale'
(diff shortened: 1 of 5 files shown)


mpinter:
^ the copilot comments above are often legit, but are all regarding logic that was already there and I don't want to make additional changes as part of this. But if any of it is of interest, the team should take notice 🙂

radoslavzeman (approved):
I don't see any major problem. 👍 

Copilot comments are definitely valid and should be revisited.

utils/languageDetectorPlugin.ts
```diff
@@ -11,12 +13,16 @@ export const languageDetectorPlugin = {
...
+      if (currentUser) {
+        const userLocale = await getAndChangeUserLanguage()
 
-      if (userLocale) {
-        return callback(userLocale)
+        if (userLocale) {
+          return callback(userLocale)
+        }
       }
     }
```
filopav, line 27: Just keep in mind that the `const { isInternetReachable } = await NetInfo.fetch()` returns false negatives many times when opening app from background, hopefully the detect doesn't run when this issue occures.

filopav (approved):
It's fine by me as well

utils/languageDetectorPlugin.ts
mpinter: :+1: good info. Not to go down this rabbit hole right now, even if this happens it would mean that the app will not sync language with backend in those cases, which should be a small enough bug/edge case.

PR #838, opened 2026-05-07
Open the additionalInformation of UDR automatically

shegood:
When opening the purchase screen of a ticket, show the additionalInformation about the UDR if there is any. 

Idea of this fix is that the default snapPoint for the `BottomSheet` component is `0` and the default value for `snapPoints` is `[HANDLE_HEIGHT]`.  When we want to expand the `BottomSheet` we need to wait for the height of the component to be calculated and appended into `snapPoints`. 

Tried a lot of other approaches, this is the only one that worked due to the `useEffects` being run before the other snapPoint is calculated

### DIFF
--- a/components/tickets/PurchaseBottomSheet.tsx
+++ b/components/tickets/PurchaseBottomSheet.tsx
@@ -37,6 +37,7 @@ const PurchaseBottomSheet = forwardRef<BottomSheet, Props>(
     return (
       <BottomSheet
         ref={ref}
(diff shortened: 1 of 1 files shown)


filopav (approved):
This approach is the correct one, not the one with use effects. The only thing is that if user hides it and moves to different udr it will get opened every times (can be a pain on smaller devices. Check it with @zuzanaslobodova and make sure it's all right from product standpoint

